PR #6, opened 2015-10-22
Fixed lastname and empty list exception

salixzs:
There were copy-paste bug with first_name and last_name picking.
Also FindName throwed exception on non-English locales due to lack of suffixes/postfixes defined.
Funny that unit test checked first name and last name the same. Now should be OK.

### DIFF
--- a/Source/Bogus.Tests/NameTests.cs
+++ b/Source/Bogus.Tests/NameTests.cs
@@ -48,6 +48,13 @@ namespace Bogus.Tests
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


bchavez:
Awesome! Thank you so much @salixzs for your contribution. Soon as CI passes I'll merge push to nuget.

bchavez:
[v3.0.1](https://www.nuget.org/packages/Bogus/) :rocket: pushed on NuGet with your fix.

Thanks again :+1:

PR #23, opened 2016-03-29
fix error: System.IndexOutOfRangeException in Commerce.Department

arumata:
You can get exception if you try this:

```
public class TestClass
{
	public string Value { get; set; }
}
[Test]
public void TestFaker()
{
	var faker = new Faker<TestClass>();
	faker.RuleFor(x => x.Value, faker1 => faker1.Random.Word());
	foreach (var item in faker.Generate(1000))
	{
		Debug.WriteLine(item.Value);
	}
}
```

### DIFF
--- a/Source/Bogus/DataSets/Commerce.cs
+++ b/Source/Bogus/DataSets/Commerce.cs
@@ -14,11 +14,14 @@ namespace Bogus.DataSets
             var num = max;
             if( !returnMax )
             {
-                 num = this.Random.Number(max);
+                do
+                {
+                    num = this.Random.Number(max);
+                } while (num == 0);
             }
 
             var cats = Categories(num);
-            if( num > 1 )
+            if (num > 1)
             {
                 return string.Format("{0} & {1}", string.Join(", ", cats.Take(cats.Length - 1)),
                     cats.Last());


Source/Bogus/DataSets/Commerce.cs
```diff
@@ -14,11 +14,14 @@ public string Department(int max = 3, bool returnMax = false)
             var num = max;
             if( !returnMax )
             {
-                 num = this.Random.Number(max);
+                do
+                {
+                    num = this.Random.Number(max);
+                } while (num == 0);
```
bchavez: Hey, there! Thanks very much for your PR and bringing this up to my attention. I really appreciate it. You're absolutely correct, I can reproduce the the exception with `Generate(1000)`. I'm not sure a `do...while` loop is best here though. I think it can be simplified to: `num = this.Random.Number(1, max);`. The `Random.Number(` overload supports a minimum parameter.  The random value we received was zero, which ultimately selected zero categories.

bchavez:
Hi @arumata , I fixed this in 88f080db1557112d609e603ed0b1c761b5f3fc7f `Number(1,num)` without the `do...while` loop.

Again, thanks for bringing up the issue. I really appreciate it.

PR #101, opened 2017-11-04
Better Determinism and Faker.Clone()

### LINKED ISSUE #100: Scoped / Thread safe verion of Randomizer.Seed = new Random(SEED);
Hi there. I saw that there's a static property which can be used for deterministic results.

```csharp
int SEED = 12345;
Randomizer.Seed = new Random(SEED);
```

However, if I use this approach with multiple threads, I get inconsistent results because my threads are asyncronous.  Is there a method that I'm overlooking which would allow me a locally scoped version of this property?  I'd very much like to be able to have 100+ concurrent operations going on and still get the benefit of determinism. 

Thanks for your time!  Love the library either way 👍

bchavez:
Resolves #100.

This PR resolves a slowly brewing code smell in **Bogus**. My careless usage of `static` variables and sporadic sprinklings of `new Randomizer()` variables as a randomness source started to surface in reports about multi-threading issues and ultimately became a roadblock for adding new features like #100.

This PR resolves the matter by making determinism, as a concept, a first-class concern inside **Bogus**.

The way determinism is implemented in this PR is by keeping track of any dependent instances of data sources, at all levels (`DataSet`, `Facade`, and `Faker<T>`) that draw from a `Randomizer`. The overall architecture looks something like:

```
//Data sets with no other sources except Randomizer.
new Name
    |
    Randomizer

//Data sets that depend on other data sets as sources
//including their own randomizer
new Internet ->  new Name  
    |            |
    Randomizer   Randomizer
        \- Notifier -->^  


//The Fake facade, the collection of all data sets
new Faker()
    |   |
    |   new Company
    |   |   |
    |   |   Randomizer
    |   |    
    |   new Address -> new Name
    |       |              |
    |       Randomizer     Randomizer
    |   ...    /  \- Notifier -->^
    |         / 
    \--Randomizer
        

//The Faker<T> rule maker
new Faker<T>()  
|          |
|          new Faker()
|              |   |
|              |   new Company
|              |   |   |
|              |   |   Randomizer
|              |   |    
|              |   new Address -> new Name
|              |       |              |
|              |       Randomizer     Randomizer
|              |   ...   /   \- Notifier -->^
|              |        /
\.UseSeed()    \--Randomizer
       \- Notifier -->^
```
Roughly, all three levels (`DataSet`, `Facade` and `Faker<T>`) have a directed acyclic dependency graph of data sources that they draw from. Anytime a new `Randomizer` property is set (at any level, `DS`, `F`, `F<T>`), the object's `Random` property contains logic to set the local instance of `Randomizer` and, *additionally*, notifies all direct-descendent child data sources downstream of the new `Randomizer`. On the receipt of the new `Randomizer`, children's direct-children are subsequently notified. Basically, the new `Randomizer` is broadcasted to all descendants of the root object.

The consequence of keeping track of what children need to be notified of the new `Randomizer` introduces some maintenance code to make sure `Randomizer` broadcasts are reached by descendants. For example, the `Internet` data set, depends on the `Name` data set to make fake data. So, the `Internet` data set constructor has the following code:
```
public Internet(string locale = "en") : base(locale)
{
    this.Name = this.Notifier.Flow(new Name(locale));
    ....
}
```
The `Notifier` is a special property of type `SeedNotfier<T>` and exists on any object that needs to track child data sources so that they are notified. The `Flow` method is just a sieve to make it as easy as possible to track objects in one line of code. This does mean we have to be a bit more concerned about how we deal with child dependencies, but that's part of making determinism a first-class citizen.

In summary, guidelines for the future and notes to myself:
* Any dependent data sources need to flow through a `Notifier`.
* Constructor code, at all levels, is prohibited from drawing values from the `Randomizer`. Drawing values from the `Randomizer` inside a constructor can break determinism. For example, it is expected that the user will probably set a random seed after the call to a constructor finishes. So, drawing randomness early inside a constructor can jack things up.

Lastly, this PR contains `Faker<T>.Clone()` feature that lets you clone the internal state of a `Faker<T>` into a new object. The following is now possible:

```csharp
public class CloneTests : SeededTest
{
   [Fact]
   public void can_create_a_simple_clone()
   {
      var orderFaker = new Faker<Examples.Order>()
         .UseSeed(88)
         .RuleFor(o => o.OrderId, f => f.IndexVariable++)
         .RuleFor(o => o.Quantity, f => f.Random.Number(1, 3))
         .RuleFor(o => o.Item, f => f.Commerce.Product());

      var clone = orderFaker.Clone();

      var clonedOrder = clone.Generate();

      var rootOrder = orderFaker.Generate();

      clonedOrder.ShouldBeEquivalentTo(rootOrder);
   }

   [Fact]
   public void clone_has_different_rules()
   {
      var rootFaker = new Faker<Examples.Order>()
         .UseSeed(88)
         .RuleFor(o => o.OrderId, f => f.IndexVariable++)
         .RuleFor(o => o.Quantity, f => f.Random.Number(1, 3))
         .RuleFor(o => o.Item, f => f.Commerce.Product());

      var cloneFaker = rootFaker.Clone()
         .RuleFor(o => o.Quantity, f => f.Random.Number(4, 6));

      var rootOrder = rootFaker.Generate();
      var clonedOrder = cloneFaker.Generate();

      rootOrder.Quantity.Should()
         .BeGreaterOrEqualTo(1).And
         .BeLessOrEqualTo(3);

      clonedOrder.Quantity.Should()
         .BeGreaterOrEqualTo(4).And
         .BeLessOrEqualTo(6);
   }
}
```

Pretty kick-ass. :+1:

:zap: :runner: ***["I was struck by lighting. Walkin' down the street..."](https://www.youtube.com/watch?v=iypUpv9xelg)***

/cc @oliverkane

### DIFF
--- a/Source/Bogus.Tests/Bogus.Tests.csproj
+++ b/Source/Bogus.Tests/Bogus.Tests.csproj
@@ -79,6 +79,7 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="CloneTests.cs" />
     <Compile Include="DataSetTests\AddressTest.cs" />
     <Compile Include="DataSetTests\CardTests.cs" />
     <Compile Include="DataSetTests\CommerceTest.cs" />
--- a/Source/Bogus.Tests/Bogus.Tests.csproj
+++ b/Source/Bogus.Tests/Bogus.Tests.csproj
@@ -90,6 +90,7 @@
     <Compile Include="ExtensionTests\ExtensionTest.cs" />
     <Compile Include="FluentTests.cs" />
     <Compile Include="GitHubIssues\Issue10.cs" />
+    <Compile Include="GitHubIssues\Issue100.cs" />
     <Compile Include="GitHubIssues\Issue13.cs" />
     <Compile Include="GitHubIssues\Issue23.cs" />
     <Compile Include="GitHubIssues\Issue45.cs" />
--- a/Source/Bogus.Tests/CloneTests.cs
+++ b/Source/Bogus.Tests/CloneTests.cs
@@ -1,0 +1,50 @@
+﻿using FluentAssertions;
+using Xunit;
+
+namespace Bogus.Tests
+{
+   public class CloneTests : SeededTest
+   {
+      [Fact]
+      public void can_create_a_simple_clone()
+      {
+         var orderFaker = new Faker<Examples.Order>()
+            .UseSeed(88)
+            .RuleFor(o => o.OrderId, f => f.IndexVariable++)
+            .RuleFor(o => o.Quantity, f => f.Random.Number(1, 3))
+            .RuleFor(o => o.Item, f => f.Commerce.Product());
+
+         var clone = orderFaker.Clone();
+
+         var clonedOrder = clone.Generate();
+
+         var rootOrder = orderFaker.Generate();
+
+         clonedOrder.ShouldBeEquivalentTo(rootOrder);
+      }
+
+      [Fact]
+      public void clone_has_different_rules()
+      {
+         var rootFaker = new Faker<Examples.Order>()
+            .UseSeed(88)
+            .RuleFor(o => o.OrderId, f => f.IndexVariable++)
+            .RuleFor(o => o.Quantity, f => f.Random.Number(1, 3))
+            .RuleFor(o => o.Item, f => f.Commerce.Product());
+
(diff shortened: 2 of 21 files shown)


oliverkane:
I just want to comment that @bchavez is a pretty responsive guy to publicly posted issues. He had no obligation to me making a request from the wild, but he did it not only quickly, but with enthusiasm and with a super slick result.  

Awesome work man!

PR #108, opened 2017-11-19
Bool random generator

supix:
Bool() method allows to specify probability to get a true value.


bchavez:
Grazie mille Marcello. Merged. :+1:

supix:
Grazie a te!

Issue #114, opened 2017-12-14
Credit Cards can be obfuscated and only get the last four digits.


rjcarneiro:
It would be nice generate CreditCardNumbers obfuscated, **** **** **** 4453 and also only the 4 last digits.

```
public System.String CreditCardNumber(CardType provider = null);
public System.String CreditCardNumber(bool obfuscate = false);
public System.String CreditCardNumber(CardType provider = null, bool obfuscate = false);
```

```
public System.String CreditCardNumberLastFourDigits(CardType provider = null);
```

bchavez:
Hi @rjcarneiro ,

Thank you for posting your issue. I appreciate it. I'm always happy to see how devs are using **Bogus**.

However, I feel a little hesitant about the idea of obfuscating credit card numbers inside **Bogus**. 

My concerns are:

* Credit card obfuscation can vary widely across apps.
   * Some might use ` ` spaces; others might use `-`, while others might not use any separators.
   * Some might apps chose to reveal a combination of first 3 and last 4.
* The last group of digits of some credit cards can end with 2, 3, or 4. IE: `****-12`, `****-123`, `****-1234`.

Personally, I feel there's too much variability in the way credit card numbers are masked in apps to really hard-code this formatting into **Bogus**. My gut instinct here is to forward any obfuscation requrement to `ReplaceNumber`:

```
f.Random.ReplaceNumbers("****-****-****-####")
// "****-****-****-9841"
```

Using `ReplaceNumbers` avoids a lot of the concerns I outlined above.

Lastly, if anyone else would like to chime in about adding this to **Bogus** feel free to comment. I'd like to know what others think too.

:hourglass_flowing_sand: :mag: ***["But I still haven't found what I'm for..."](https://www.youtube.com/watch?v=wdCJRybAtso)***

rjcarneiro:
Hi @bchavez ,

Thanks for your reply. 

I believe this new features can improve your package setting the framework more capable. Your concerns are valid and it's possible to create more overloads to solve that.

bchavez:
Hi @rjcarneiro ,

I apologize for taking so long to get your code merged. I promoted your credit card obfuscation methods to **Extension Methods** and moved them from the core `Finance` dataset class to the `Bogus.Extensions.Extras` namespace.

I hope you feel this is okay. I think this is a good compromise for both of us. As I mentioned earlier about my concerns I'm okay letting them live in an extension namespace; just maybe not so much in the core API of `Finance` dataset class.

[...]

PR #118, opened 2017-12-16
Credit card obfuscator and last four digits

rjcarneiro:
Related with #114

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
-
 ## v22.0.1
 * Issue 120: `.Generate(n)` now returns `List<T>` instead of `IList<T>`.
 * Added `f.Address.CountryCode()` ISO 3166-1 alpha-3 country code generator.
+* Added `Bogus.Extensions.Extra` namespace for generally useful helper methods.
 
 ## v21.0.5
 * Better error support.
--- a/Source/Bogus.Tests/Bogus.Tests.csproj
+++ b/Source/Bogus.Tests/Bogus.Tests.csproj
@@ -96,6 +96,7 @@
     <Compile Include="GitHubIssues\Issue102.cs" />
     <Compile Include="GitHubIssues\Issue104.cs" />
     <Compile Include="GitHubIssues\Issue115.cs" />
+    <Compile Include="GitHubIssues\Issue114.cs" />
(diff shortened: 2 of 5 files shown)


bchavez:
Hi @rjcarneiro ,

Thank you for taking the time to create a PR for your issue. I really apprecate your effort and motivation here; however...

I still feel masking credit card numbers is a specific UI/app formatting concern. I really feel bad rejecting PRs but this PR hard-codes a specific masking format in C#.

Let me try to explain why I think this is the wrong approch:

What if the users of your `CreditCardNumberObfuscated` method wanted the first 4 and last 4 digits of the obfuscated number? You can imagine another parameter overload to `CreditCardNumberObfuscated` to handle that specific use-case. And on, and on, and on... This is why I don't think `CreditCardNumberObfuscated` is a mintainable approch to the problem of obfuscating credit card numbers. It just doesn't sit well with me for some reason.

I'd still recommend developers who require credit card obfuscation to use `ReplaceNumbers` manually where they can specifiy any format they want: `Random.ReplaceNumbers($"****what-ever-format-I-want-####");`. I'm not sure `CreditCardNumberObfuscated` serving as a shortcut adds much value over what they can do manually except avoid the hastle of typing-out a format. If typing-out an obfuscated format was a hastle a custom extension method can easily serve the same purpose of `CreditCardNumberObfuscated` without being included in **Bogus**.

I hope that makes sense.

:dash: :walking: ***["Bubbles of gas in my brain... Send me off balance, it's not enough"](https://www.youtube.com/watch?v=AS6NfeREKPk)***

rjcarneiro:
Hi @bchavez, 

Thanks for your reply.

My main motivation to develop this methods were to get more accurate date for [PCI Compliance](https://www.pcicomplianceguide.org/faq/).

There are rules, standards and etics about developing PCI application. Things like:

- when a card is obfuscated only the last four digits are shown;
- never show, log or print the full credit card number;

And the list goes on... In a PCI world, you never show the first four dgitis because they tell which type of card you have. 

For me, this methods will be really helpful for other develops read the code, understand and get more accurate data into test and integrations.

However, if you feel that is not appropriate, feel free to refuse the pull request created by me.

Cheers

PR #136, opened 2018-03-29
Improve speed of DataSet.ParseTokens() 

danij:
Profiling an application that uses Bogus to generate thousands of entities indicates a lot of time being spent recreating the regular expression whenever `DataSet.ParseTokens()` is called.

By caching and compiling the regular expression, performance should improve considerably. According to the [official documentation](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex(v=vs.110).aspx):

> The Regex class is immutable (read-only) and thread safe. Regex objects can be created on any thread and shared between threads.

### DIFF
--- a/Source/Bogus/DataSet.cs
+++ b/Source/Bogus/DataSet.cs
@@ -143,14 +143,15 @@ namespace Bogus
          return Random.Replace(tokenResult);
       }
 
+      private static readonly Regex parseTokensRegex = new Regex("\\#{(.*?)\\}", RegexOptions.Compiled);
+
       /// <summary>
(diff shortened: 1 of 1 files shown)


bchavez:
[**Bogus** `v22.0.6`](https://www.nuget.org/packages/Bogus/22.0.6) has been published with your fix. :)

Thanks again,
Brian

:chocolate_bar: :cookie: :lollipop: [**Ronald Jenkees - Stay Crunchy**](https://www.youtube.com/watch?v=GeIAXlwVlZc)

PR #151, opened 2018-06-24
Generate random numbers based on a Normal Distribution (Gaussian)

codersg:
Bogus already has excellent methods for generating random numbers, but these numbers are all, by design, evenly distributed across the whole range of min to max.

In some cases, this even distribution will result in generating data that is not realistic. As an example, if we wanted to generate random data to be used for “Height of a person”, we would currently get an equal number of short, average and tall people being generated. But if we know so properties of the data we want to generate (for example, the US the average height of a male is 175.7cm with a standard deviation of 10cm), it would be great to be able to generate random data points that when analysed, matched the average and standard deviation, meaning we’d have far less extremely short and extremely tall people, which would reflect realistic data.

This could also be used for fields such as product prices, ages and so forth. It’s often more common to have most values falling with a certain range, and less items having extreme values as the tail of the min and max.

I think it would make Bogus even more useful to have this ability. although Bogus is not designed for generating statistical data, the addition of this method is very simple and gives the option of much more realistic data being generated if it's required.

Therefore, I’ve added some methods, prefixed with “FromNormalDistribution” to allow you to specify the mean and standard deviation instead of a min and max.

This is my first ever pull request, so if I'm doing anything wrong, please let me know! Always willing to learn and do this the correct and polite way. :)

### DIFF
--- a/Source/Bogus/Randomizer.cs
+++ b/Source/Bogus/Randomizer.cs
@@ -121,7 +121,76 @@ namespace Bogus
          return result;
       }
 
+      // Coefficients used in Acklam's Inverse Normal Cumulative Distribution function.
+      private static readonly double[] AklmasCoefficientA =
+         {-39.696830d, 220.946098d, -275.928510d, 138.357751d, -30.664798d, 2.506628d};
 
+      private static readonly double[] AklamsCoefficientB =
+         {-54.476098d, 161.585836d, -155.698979d, 66.801311d, -13.280681d};
+      private static readonly double[] AklamsCoefficientC =
+         {-0.007784894002d, -0.32239645d, -2.400758d, -2.549732d, 4.374664d, 2.938163d};
+
+      private static readonly double[] AklamsCoefficientD = { 0.007784695709d, 0.32246712d, 2.445134d, 3.754408d };
+
+      // Break-Points used in Aklam's Inverse Normal Cumulative Distribution function.
+      private const double AklamsLowBreakPoint = 0.02425d;
+      private const double AklamsHighBreakPoint = 1.0d - AklamsLowBreakPoint;
+
+      /// <summary>
+      /// This algorithm follows Peter J Acklam's Inverse Normal Cumulative Distribution function.
+      /// Reference: P.J. Acklam, "An algorithm for computing the inverse normal cumulative distribution function," 2010
+      /// </summary>
+      /// <param name="probability"></param>
+      /// <returns>
+      /// A double between 0.0 and 1.0
+      /// </returns>
+      private static double InverseCumulativeStandardNormalDistribution(double probability)
+      {
+
+         // Rational approximation for lower region of distirbution
+         if (probability < AklamsLowBreakPoint)
+         {
+            double q = Math.Sqrt(-2 * Math.Log(probability));
+            return (((((AklamsCoefficientC[0] * q + AklamsCoefficientC[1]) * q + AklamsCoefficientC[2]) * q + AklamsCoefficientC[3]) * q +
+                     AklamsCoefficientC[4]) * q + AklamsCoefficientC[5]) /
+                   ((((AklamsCoefficientD[0] * q + AklamsCoefficientD[1]) * q + AklamsCoefficientD[2]) * q + AklamsCoefficientD[3]) * q + 1);
+         }
+
+         // Rational approximation for upper region of distribution
+         if (AklamsHighBreakPoint < probability)
+         {
+            double q = Math.Sqrt(-2 * Math.Log(1 - probability));
+            return -(((((AklamsCoefficientC[0] * q + AklamsCoefficientC[1]) * q + AklamsCoefficientC[2]) * q + AklamsCoefficientC[3]) * q +
+                      AklamsCoefficientC[4]) * q + AklamsCoefficientC[5]) /
+                   ((((AklamsCoefficientD[0] * q + AklamsCoefficientD[1]) * q + AklamsCoefficientD[2]) * q + AklamsCoefficientD[3]) * q + 1);
+         }
+
+         // Rational approximation for central region of disribution
+         {
+            double q = probability - 0.5d;
+            double r = q * q;
+            return (((((AklmasCoefficientA[0] * r + AklmasCoefficientA[1]) * r + AklmasCoefficientA[2]) * r + AklmasCoefficientA[3]) * r +
+                     AklmasCoefficientA[4]) * r + AklmasCoefficientA[5]) * q /
+                   (((((AklamsCoefficientB[0] * r + AklamsCoefficientB[1]) * r + AklamsCoefficientB[2]) * r + AklamsCoefficientB[3]) * r +
+                     AklamsCoefficientB[4]) * r + 1);
+         }
+      }
+
+      /// <summary>
+      /// Generate a random double, based on the specified normal distribution.
+      /// </summary>
+      /// <param name="mean">Mean value of the normal distribution</param>
+      /// <param name="standardDeviation">Standard deviation of the normal distribution</param>
+      public double DoubleFromNormalDistribution(double mean, double standardDeviation)
+      {
+         //lock any seed access, for thread safety.
+         lock (Locker.Value)
+         {
+            double p = InverseCumulativeStandardNormalDistribution(localSeed.NextDouble());
+            return (p * standardDeviation) + mean;
+         }
+      }
+      
       /// <summary>
       /// Get a random double, between 0.0 and 1.0.
       /// </summary>
@@ -153,6 +222,16 @@ namespace Bogus
       }
 
       /// <summary>
+      /// Generate a random decimal, based on the specified normal distribution.
+      /// </summary>
+      /// <param name="mean">Mean average of the normal distribution</param>
+      /// <param name="standardDeviation">Standard deviation of the normal distribution</param>
+      public decimal DecimalFromNormalDistribution(double mean, double standardDeviation)
+      {
+         return Convert.ToDecimal(DoubleFromNormalDistribution(mean, standardDeviation));
+      }
+
+      /// <summary>
       /// Get a random float, between 0.0 and 1.0
       /// </summary>
       /// <param name="min">Minimum, default 0.0</param>
@@ -160,6 +239,16 @@ namespace Bogus
       public float Float(float min = 0.0f, float max = 1.0f)
       {
          return Convert.ToSingle(Double()) * (max - min) + min;
+      }
+
+      /// <summary>
+      /// Generate a random float, based on the specified normal distribution.
+      /// </summary>
+      /// <param name="mean">Mean average of the normal distribution</param>
+      /// <param name="standardDeviation">Standard deviation of the normal distribution</param>
+      public float FloatFromNormalDistribution(double mean, double standardDeviation)
+      {
+         return Convert.ToSingle(DoubleFromNormalDistribution(mean, standardDeviation));
       }
 
       /// <summary>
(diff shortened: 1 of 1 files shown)


bchavez:
Hi Rob,

First off, congrats on your first **Pull Request**! It's really awesome receiving pull requests from our community especially when it's their first **Pull Request**! So, congrats and thanks! :sunglasses: 

Just a quick **pro tip** for your future endeavors here on GitHub: it's usually a good idea to create an **Issue** first before sending a **Pull Request**. When you create an **Issue** first, it gives everyone watching the repository a chance to discuss your **Issue**. It's also a great way to start a conversation about the proposed changes you're looking to make. Nothing hurts more than spending hours on a **Pull Request**, then have it ultimately rejected by the maintainer. And as a maintainer, nothing sucks more than rejecting someone's hard work. :cry:  It does happen, so don't let it get you down. I've been on both sides but a lot can be avoided by creating an **Issue** first and starting a conversation. :+1: 

----

Now, let's talk about your **Pull Request**! You outline a compelling use case. The idea of Bogus supporting different statistical distributions has been in the back of my mind. I haven't acted on any implementation thus far because 1) I didn't need it and 2) nobody has requested it yet until now. I do think Bogus should support some kind of statistical distribution story for cases you've outlined above but what form does it take? I don't know yet and I'd like to explore this idea a little more.

* Are there any other examples besides being in the set of real numbers that could benefit from statistical distributions? Like generating email addresses from a particular domain that's more popular than others? It would really help me understand the scope and get a feel for the generalization/usefulness of distributions. If you could provide any real-world examples you've encountered during software development (excluding the set of real numbers) when creating fake data that would be great.

* What are your thoughts about moving your implementation into a dedicated namespace as a set of extension methods to `Randomizer`?

  So, instead of:
  ```csharp
  var faker = new Faker();
  var height = faker.Random.DoubleFromNormalDistribution(mean:_, standardDeviation:_);
  ``` 
  We have something like:
  ```csharp
  using Bogus.Distributions.Gaussian;

  var faker = new Faker();
  var height = faker.Random.GuassianDouble(mean:_, standardDeviation:_);

  using Bogus.Distributions.Poisson;
  var hondaCars = faker.Random.PoissonInt(_)
  ```
  It feels like `Random.GaussianDouble(_)`, `Random.GaussianFloat(_)`, and `Random.GaussianInt(_)` are a little more approachable and ergonomic than `Random.*FromNormalDistribution(...)` methods. Also, I think using extension methods opens up the possibility of supporting other probability distributions without overpopulating `Randomizer`'s core methods. Ideally, I'd like to keep Bogus' core APIs simple and approachable for 90% of devs and use cases; but for the other 10% that want to push Bogus further with more advanced capabilities, it should be as easy as flipping a switch to light up new advanced APIs.
 
  I'm very much open to different namespaces for this kind of probability code. Some that come to mind:
    * `using Bogus.Distributions.Gaussian`
    * `using Bogus.Math.Distributions.Gaussian`
    * `using Bogus.Probability.Distributions.Gaussian`
    * `using Bogus.Extensions.Math.Distributions.Gaussian`

* Is it possible to write some unit tests for your **Gaussian** normal distribution implementation? It would be great if your PR could include some unit tests to make sure we have this implementation locked down in case of any future changes to Bogus or your implementation.

So, let me know your thoughts on the points above. You've done some really great work here in getting Bogus started down the path of supporting probability distributions! Let's work on getting your code merged!

Thanks,
Brian

:walking_man: :walking_woman: **[Missing Persons - Walking in L.A.](https://www.youtube.com/watch?v=R_UpLtGEWoY)**

Source/Bogus/Randomizer.cs
```diff
@@ -152,6 +221,16 @@ public decimal Decimal(decimal min = 0.0m, decimal max = 1.0m)
          return Convert.ToDecimal(Double()) * (max - min) + min;
       }
 
+      /// <summary>
+      /// Generate a random decimal, based on the specified normal distribution.
+      /// </summary>
+      /// <param name="mean">Mean average of the normal distribution</param>
+      /// <param name="standardDeviation">Standard deviation of the normal distribution</param>
+      public decimal DecimalFromNormalDistribution(double mean, double standardDeviation)
```
bchavez: Also, it would be really helpful if you could include an example of the typical usage of your method. Just so people don't go off throwing numbers crazy nubmers as parameters. Something like:
```
/// <summary>
/// Generate a random decimal, based on the specified normal distribution.
/// For example, to create random values around an average height of 69.1
/// inches with a standard deviation of 2.9 inches away from the mean
/// call DecimalFromNormalDistribution(69.1, 2.9).
/// </summary>
/// <param name="mean">Mean average of the normal distribution</param>
/// <param name="standardDeviation">Standard deviation of the normal distribution</param>
```
I usually find it really helpful when XML docs have a small example use of how to use the method.

codersg:
Hi Brian,

Many thanks for the feedback, good tips and for pointing me in the right direction with GitHub etiquette and norms. It helps set my expectations and helps me begin my journey of being a contributor to community projects. 😄 

Regarding the pull request, some of my thoughts of your questions:

**> Are there any other examples besides being in the set of real numbers that could benefit from statistical distributions?**

Yes, I think lots of non-numerical data could benefit from having a defined probably of being chosen. But it will mean having to specify a probability for each discrete value. An example might be the Web Browser” used in some generated logs. According to Wikipedia, latest data suggests that for Tablet browsers, 58% of traffic is from Chrome, 35% from Safari, 4% from AOSP and the remainder from other browsers.

The user would need to be able to specify the percentages for each value, or perhaps just some of them and allow the remaining values to be have an equal probably of being chosen among the remaining probability. For example, you could just specify “Chrome @ 58%” and let all the other values be chosen with an equal chance , 42% of the time.

I could potentially look at adding something like this. I guess this wouldn't fall under the Gaussian extensions though?

> **What are your thoughts about moving your implementation into a dedicated namespace as a set of extension methods to Randomizer?**

That method names sounds much better! I always tend to name methods and variables things a bit long 😊 

Personally I like your suggest of using the `namespace Bogus.Distributions.Gaussian`, as I think the use of `Math` might put some people off (like me) and the others seemed a bit long. This namespace is clear (in my opinion) and would allow for other Distributions to be implemented.

> **Is it possible to write some unit tests for your Gaussian normal distribution implementation?** 

It should be, but as we’re dealing with random numbers and probability, it’s possible (just unlikely, but not incorrect) that the generated data wouldn’t fit exactly a specified statistical characteristic, so might fail. But given enough samples, I guess it should always follow some pattern with a tolerance. 

And noted on including examples in the XML docs. You're right, its always good to have some small documentation or example in the code itself, instead of having to go look it up in the documentation.

bchavez:
Hi Rob,

Cool. Sounds good. You can go ahead and make the changes.

* Create a folder in `Bogus\Distributions\Gussian\` add a file `ExtensionsForRandomizer.cs` inside the new folder.
* Move your code into the extension namespace and refactor your methods with the `Gaussian*` prefix.
* Update the XMLDocs with a basic/practical usage example.
* Add a new unit test in `Bogus.Tests\DistributionTests\Gussian.cs` that tests your new extension methods. It doesn't have to be super crazy, just something that captures the essence of functionality and makes a big noise if something is glaringly out of range or bounds.
* Commit the changes on top of what you currently have in your repo.
* Push those changes to your fork on GitHub. Since you already have an open PR your new changes will automatically appear here as an "update" to your current PR.

We'll do one final review, and if it all looks good, we'll merge. :+1: 

> I could potentially look at adding something like this. I guess this wouldn't fall under the Gaussian extensions though?

Nah, it's cool. You don't have to right now. Keeping PR's focused and somewhat small & digestible is good. I just wanted to get some idea about how you used distributions in your software for fake data and get a sense of the kind of integration that would be necessary to make Bogus work well for such use cases.

I think what you have right now is perfect and a really good start. We just need to move it in a new namespace and refactor the methods! :sunglasses: 

Thanks,
Brian

codersg:
Hi Brian.

I’ve made the changes as requested; when you get a chance would you mind
reviewing them and checking whether it’s what you were expecting and
whether it’s sufficient enough. The comments feel a little repetitive, but
they’re essentially the same method just with different data types returned.

It’s also my first time using xUnit (lots of firsts for me here), but the
existing examples were very quite clear, so I managed to work out how it
works using those as examples. I’ve added only one test per data type just
to check the fundamental nature of these methods, i.e. produce random data
that when analysed matches the specified mean and standard distribution (to
a certain tolerance). I allowed a 10% tolerance although I was surprised to
see it’s actually far more accurate than that, which I wasn’t expecting;
but still I think a 10% tolerance is a reasonable expectation, so have left
it in there at this value. Maybe a true statistician will look at it one
day and do a better effort. 😊

I look forward to your feedback; and thanks again for the pointers.

Rob.

bchavez:
Hi Rob!

First off, great work! There is some real computer science going on in that code! :+1: Code comments in unit tests are great! XML docs are looking perfect! Very nice!

I only have one observation: What do you think about changing the naming convention of the extension methods to `.GaussianType()` instead of `.TypeGaussian()`? I'm thinking about API discoverability for the extension methods. Currently, we have:

```csharp
using Bogus.Distributions.Gaussian;

var f = new Faker();
f.Random.DoubleGaussian(...);
f.Random.IntGaussian(...);
f.Random.FloatGaussian(...);
f.Random.DecimalGaussian(...);

var fakeOrder = new Faker<Product>()
                 .RuleFor(o => o.Price, f => f.Random.DecimalGaussian(...) )
                 .RuleFor(o => o.Rating1, f => f.Random.FloatGaussian(...) )
                 .RuleFor(o => o.Rating2, f => f.Random.FloatGaussian(...) );
```
Let's say a developer wants to use your great Gaussian extension methods; so they import `using Bogus.Distributions.Gaussian` at the top of their code file. Right off the bat, we know they are after those Gaussian extension methods. As soon as they press `f.Random.` they'll be presented with a relatively large list of methods to choose from by IntelliSense. If the Gaussian extension methods aren't grouped together I think we'd miss an important opportunity to show the developer what Gaussian extension methods lit up as a result of adding the Gaussian namespace. Also, I think Visual Studio, VSCode, Resharper IntelliSense, and Rider IDE, would have `.IntGaussian()` and `.DoubleGussian()` spaced pretty far apart away from each other if the method list was in alphabetical order.

Also, when you're writing code focused on Gaussian stuff, you'd probably want to know what new superpowers you gained (without looking at documentation) as soon as you typed: `f.Random.Gau`.

Also, if we stick to a method naming convention for `Bogus.Distributions.Kind` that follows `.KindType()` we could have additional extension methods that extends datasets as well. And it would be realitvely easy for the developer to discover all the Gaussian stuff that lights up in different places across Bogus because developers would be acustom to the `.Gaussian__` method prefix that would light up everywhere (not just in `Randomizer`). Imagine `.PoissonType()` and all the places it would extend too.

Also, maybe the code reads a little better?
```csharp
using Bogus.Distributions.Gaussian;

var f = new Faker();
f.Random.GaussianDouble(...);
f.Random.GaussianInt(...);
f.Random.GaussianFloat(...);
f.Random.GaussianDecimal(...);

var fakeOrder = new Faker<Product>()
                 .RuleFor(o => o.Price, f => f.Random.GaussianDecimal(...) )
                 .RuleFor(o => o.Rating1, f => f.Random.GaussianFloat(...) )
                 .RuleFor(o => o.Rating2, f => f.Random.GaussianFloat(...) );
```
I'd like to know what you think.

I hope I don't come off too nitpicky, but sometimes I find it's just the little details and small things that make great APIs and feel right. Regardless, I think you've done a beautiful job here! :+1: 

Thanks,
Brian

:black_circle: :dizzy: ***["Black magic... bla-a-a-ack... black magic..."](https://www.youtube.com/watch?v=3SnlsTtUZK0)***

codersg:
Hi Brian

I agree; it makes much more sense to rename these, especially for the grouping of functions with IntelliSense and the general increased readability. I think I was putting too much emphasis on the datatype rather than the purpose.

I’ve renamed the methods and updated the comments accordingly.

And nope, not too nit-picky at all. It’s this kind of feedback that helps people like me improve and write more usable code, so it’s all very welcome. I also appreciate your effort of giving clear and detailed explanations of why you suggest the changes.

Thanks again,
Rob.

bchavez:
Hi Rob,

Your changes are live on **NuGet** in [`Bogus 22.1.4`](https://www.nuget.org/packages/Bogus/22.1.4)!

On behalf of the Bogus community, thank you for your contribution!

You did great work here espeically for your first PR! Wish you the best of luck in your future PRs! :sunglasses: :game_die: 

Thanks,
Brian

codersg:
Thanks Brian,

Feels good to have been able contribute in some minor way to the project.

How should I go about starting a discussion on the topic you mentioned
before, of having other types of data being generated with some form of
probability associated with them? Do I just open an issue and propose the
concept there? "Issue" sounds like the wrong term for a suggestion. ;)

PR #153, opened 2018-06-29
spelling

SimonCropp:
note there are a few breaking changes in here


bchavez:
Hi Simon!

Thanks a bunch for these wonderful fixes!

As far as I can tell, the breaking changes should be okay since most of them were parameter name changes.  I think as long as we didn't move methods around or change the deterministic output of the unit tests in a major way, we should be good. I'll make a note of the changes in the changelog.

Thanks again for your contribution, I appreciate it!

I'll have a new release out shortly.

PR #155, opened 2018-06-30
remove redundant new

SimonCropp:
isnt the new constraint redundant if a factory is required?


bchavez:
Hi Simon,

Indeed it is. Thanks again for the refinements and fixing these oddities. :+1: 

It's always appreciated. :sunglasses: 

Thanks,
Brian

:crescent_moon: :stars: ***["Nothing good happens past 2am..."](https://www.youtube.com/watch?v=VOlcDBXKhSU)***

PR #162, opened 2018-07-07
Stuff

SimonCropp:


Source/Bogus.Tests/FluentTests.cs
```diff
@@ -91,8 +91,8 @@ public void Without_Fluent_Syntax()
       public void With_Faker_Facade()
       {
          var faker = new Faker("en");
-         var o = new Order()
             {
+         var o = new Order
```
bchavez: I think these `new Order` changes are syntactically incorrect.

Source/Bogus.Tests/GitHubIssues/Issue100.cs
```diff
@@ -40,8 +39,8 @@ public void dataset_test()
       [Fact]
       public void faker_test()
       {
-         var f = new Faker()
             {
+         var f = new Faker
```
bchavez: Syntactically incorrect here too?

Source/Bogus/Distributions/Gaussian/ExtensionsForRandomizer.cs
```diff
@@ -110,9 +107,6 @@ public static int GaussianInt(this Randomizer rnd, double mean, double standardD
       /// </code>
       /// </example>
       /// </summary>
-      /// <param name="rnd"></param>
-      /// <param name="mean"></param>
-      /// <param name="standardDeviation"></param>
```
bchavez: Might be best to copy/paste the parameter XML docs from the other similar methods instead of removing.

Source/Bogus/Premium/LicenseVerifier.cs
```diff
@@ -103,8 +103,8 @@ public static void ReadLicense(string path, out string name, out string key)
          key = lines[1];
       }
 
-      public static List<string> ProbePaths { get; } = new List<string>()
          {
+      public static List<string> ProbePaths { get; } = new List<string>
```
bchavez: `CS1526 A new expression requires (), [], or {} after type`

Source/Bogus/Hashids.cs
```diff
@@ -39,9 +39,6 @@ public Hashids() : this(string.Empty, 0, DEFAULT_ALPHABET, DEFAULT_SEPS)
       /// <summary>
```
bchavez: I'd try to avoid making changes to `Hashids.cs` when possible. This file is a copy/paste from the `/Source/hashids` submodule with "a few" changes to the original source. We should avoid changes to this file as much as possible to maintain closeness to the original source in case of the unlikely event we need to update this file again.

Ultimately, though, this submodule will be deprecated in the future.

Source/Bogus/Faker[T].cs
```diff
@@ -235,8 +235,8 @@ public virtual Faker<T> Rules(Action<Faker, T> setActions)
                return null;
             };
          var guid = Guid.NewGuid().ToString();
-         var rule = new PopulateAction<T>()
             {
+         var rule = new PopulateAction<T>
```
Comment (bchavez) [inline Source/Bogus/Faker[T].cs]: Looks like an incorrect syntax here again. The statements look like they get moved below the object initializer brace.

SimonCropp:
no idea how i manage the braces stuff up. should be fixed now

Source/Bogus/Distributions/Gaussian/ExtensionsForRandomizer.cs
SimonCropp: done

Source/Bogus/Hashids.cs
SimonCropp: reverted

bchavez:
That was a lot of cruft that built up over the years. :shower:    

Looks nice! Thanks a bunch Simon! :+1:

-Brian

:briefcase: :necktie: ***["Taking care of business every day... Taking care of business every way..."](https://www.youtube.com/watch?v=mmwic9kFx2c)***

PR #164, opened 2018-07-18
New Extension for Portugal

jpassarellijr:
Generate random Tax identification number (NIF and NIPC) for Portugal. Added Facts in the ExtensionTest.

### DIFF
--- a/Source/Bogus/Extensions/Portugal/ExtensionsForPortugal.cs
+++ b/Source/Bogus/Extensions/Portugal/ExtensionsForPortugal.cs
@@ -47,8 +47,8 @@ namespace Bogus.Extensions.Portugal
          {
             return p.context[Key] as string;
          }
-
-         var id = new int[1] { NifIdentify[new Random().Next(0, NifIdentify.Length)] };
+         
+         var id =  new int[1] { NifIdentify[p.Random.Int(0, NifIdentify.Length)] };
          var digits = p.Random.Digits(7);
 
          var nifNumber = id.Concat(digits).ToArray();
@@ -67,7 +67,7 @@ namespace Bogus.Extensions.Portugal
       /// <param name="c">Object will receive the NIPC value</param>
       public static string Nipc(this Company c)
       {
-         var id = new int[1] { NipcIdentify[new Random().Next(0, NipcIdentify.Length)] };
+         var id = new int[1] { NipcIdentify[c.Random.Int(0, NipcIdentify.Length)] };
          var digits = c.Random.Digits(7);
 
          var nipcNumber = id.Concat(digits).ToArray();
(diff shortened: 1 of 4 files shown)


Source/Bogus/Extensions/Portugal/ExtensionsForPortugal.cs
```diff
@@ -0,0 +1,78 @@
...
+      /// <param name="p">Object will receive the NIF value</param>
+      public static string Nif(this Person p)
+      {
+         const string Key = nameof(ExtensionsForPortugal) + "NIF";
+         if (p.context.ContainsKey(Key))
+         {
+            return p.context[Key] as string;
+         }
+
+         var id = new int[1] { NifIdentify[new Random().Next(0, NifIdentify.Length)] };
```
bchavez: Hi @jpassarellijr, thank you very much for this pull request. :portugal: It looks like a great addition! Good job! :+1:

I do have a few concerns, could you change `new Random().Next(...` call here? Preferably, it would be `p.Random.Number(0, length - 1)` or `p.Random.Int(0, length - 1)`. Generally, the source of all randomness needs to come from an existing **Bogus** `Randomizer` managed by **Bogus** to ensure some level of sanity when it comes to providing deterministic behavior. #100 and #101 has the details if you're intrested.

When we create `new Random` objects like this, we break the deterministic behavior for users especially when there is no seed for `new Random()`. Ultimately, we need to avoid using `new Random()` objects in **Bogus** and stick to using an existing `Randomizer`.

Hope this makes sense.

Thanks,
Brian

Source/Bogus/Extensions/Portugal/ExtensionsForPortugal.cs
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      /// <summary>
+      /// NIPC - Collective Identification Number is the most correct term to refer to a company's NIF.
+      /// The first digit can be 5, 6 public collective, 8, irregular legal person or provisional number.
+      /// </summary>
+      /// <param name="c">Object will receive the NIPC value</param>
+      public static string Nipc(this Company c)
+      {
+         var id = new int[1] { NipcIdentify[new Random().Next(0, NipcIdentify.Length)] };
```
bchavez: Line here too should change: `new Random().Next(` with `c.Random.Int(0, length-1)` or `c.Random.Number(0, length-1)`. See previous comment. :+1:

jpassarellijr:
First, thank you for acknowledgement, i´m so glad!
And about the concerns, sorry! My mistake, forgot change this.
Wrote the concept but forgot change for use Randomizer component.
I´m changed now :)

bchavez:
Congrats on your first pull request @jpassarellijr ! :+1: :sunglasses: 

Your changes are now live in [**`Bogus v22.3.2`**](https://www.nuget.org/packages/Bogus/22.3.2) :tada:

Thanks again for your help!
Brian

PR #160, opened 2018-07-05
Add DateTimeOffset

SimonCropp:
fixes #158

PR #165, opened 2018-07-20
add utcnow support

### LINKED ISSUE #139: UtcNow support for Date generation
Currently dates are generated relative to `DateTime.Now`.  Is it possible to extend this to use `DateTime.UtcNow`.

SimonCropp:
fixes #139

### DIFF
--- a/Source/Bogus.Tests/DataSetTests/DateTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/DateTest.cs
@@ -315,5 +315,36 @@ namespace Bogus.Tests.DataSetTests
             .And
             .BeOnOrBefore(utcEnd);
       }
+
+
+      [Fact]
+      public void can_set_global_static_time_source()
+      {
+         Date.SystemClock = () => DateTime.UtcNow;
+
+         var d = new Date();
+
+         d.Soon().Kind.Should().Be(DateTimeKind.Utc);
+         d.Future().Kind.Should().Be(DateTimeKind.Utc);
+         d.Past().Kind.Should().Be(DateTimeKind.Utc);
+         d.Recent().Kind.Should().Be(DateTimeKind.Utc);
+
+         d.SoonOffset().Offset.Should().Be(DateTimeOffset.UtcNow.Offset);
+         d.FutureOffset().Offset.Should().Be(DateTimeOffset.UtcNow.Offset);
+         d.PastOffset().Offset.Should().Be(DateTimeOffset.UtcNow.Offset);
+         d.RecentOffset().Offset.Should().Be(DateTimeOffset.UtcNow.Offset);
+
+         Date.SystemClock = () => DateTime.Now;
+
+         d.Soon().Kind.Should().Be(DateTimeKind.Local);
+         d.Future().Kind.Should().Be(DateTimeKind.Local);
+         d.Past().Kind.Should().Be(DateTimeKind.Local);
+         d.Recent().Kind.Should().Be(DateTimeKind.Local);
+
+         d.SoonOffset().Offset.Should().Be(DateTimeOffset.Now.Offset);
+         d.FutureOffset().Offset.Should().Be(DateTimeOffset.Now.Offset);
+         d.PastOffset().Offset.Should().Be(DateTimeOffset.Now.Offset);
+         d.RecentOffset().Offset.Should().Be(DateTimeOffset.Now.Offset);
+      }
    }
 }
--- a/Source/Bogus/DataSets/Date.cs
+++ b/Source/Bogus/DataSets/Date.cs
@@ -7,21 +7,54 @@ namespace Bogus.DataSets
    /// </summary>
    public class Date : DataSet
    {
+      private Func<DateTime> dateTimeNow;
+      private Func<DateTimeOffset> dateTimeOffsetNow;
       private bool hasMonthWideContext;
       private bool hasMonthAbbrContext;
       private bool hasWeekdayWideContext;
       private bool hasWeekdayAbbrContext;
-
+      private static Func<DateTimeOffset> defaultDateTimeOffsetNow = () => DateTimeOffset.Now;
+      private static Func<DateTime> defaultDateTimeNow=() => DateTime.Now;
 
       /// <summary>
       /// Create a Date dataset
       /// </summary>
-      public Date(string locale = "en") : base(locale)
+      public Date(string locale = "en", Func<DateTime> dateTimeNow=null, Func<DateTimeOffset> dateTimeOffsetNow = null) : base(locale)
       {
+         if (dateTimeNow == null)
+         {
+            this.dateTimeNow = defaultDateTimeNow;
+         }
+         else
+         {
+            this.dateTimeNow = dateTimeNow;
+         }
+         if (dateTimeOffsetNow == null)
+         {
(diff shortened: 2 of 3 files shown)


bchavez:
Hi Simon!

Hope you are well! Thank you for taking a look at #139. I really appreciate it! It's always nice to have community help!

However, there are some thoughts I have about this PR and #139 in general. I'll try to explain my current thinking on the subject.

### The PR changes
My first thought is on the constructor changes:

```csharp
public Faker(string locale = "en",
                Func<DateTime> dateTimeNow = null,
                Func<DateTimeOffset> dateTimeOffsetNow = null){
...
}

public Date(string locale = "en",
                Func<DateTime> dateTimeNow=null,
                Func<DateTimeOffset> dateTimeOffsetNow = null) : base(locale)
       {
```
* I've avoided changing the constructor object on `DataSet`s, `Faker` and `Faker<T>` objects because I want to keep constructors as simple as possible.
* Let's imagine the API usage for a second:
   ```csharp
  //hmm.
  var f1 = new Faker(dateTimeNow: () => DateTime.UtcNow);

  //if I want to set both
  var f2 = new Faker(
     dateTimeNow: () => DateTime.UtcNow,
     dateTimeOffsetNow: () => DateTimeOffset.UtcNow);

  //and not much better. imo
  Func<DateTimeOffset> nowOffset = () => DateTimeOffset.UtcNow;
  Func<DateTime> now = () => DateTime.UtcNow;
  var f3 = new Faker(dateTimeNow: now, dateTimeOffsetNow: nowOffset);
   ```
   I don't know, there just seems like there's a lot of API noise around this constructor that doesn't seem (at first glance) much to do with the issue at hand, creating a `Faker` object. The `Date` dataset details are bubbling into the `Faker` constructor. Indeed, these constructor parameters seem more appropriate on the `Date` dataset, but less so the further we move up the abstraction ladder to `Faker` and `Faker<T>`.
* Also, I'm not sure this is a scalable approach. I can imagine, we have some dataset in the future (even currently now) that could accept some default settings. Would our approach be to add more constructor parameters for these datasets? It doesn't seem right either. I don't think adding more constructor parameters for new settings is the way to go. Things like `Randomizer.Seed` and `Faker.DefaultStrictMode` are some settings that come into mind.

#### The right approch, maybe...
At a very minimum, it seems more appropriate to collapse these "default" settings in a `Settings` object that we pass around by calling a method similar to `.UseSeed()`. Or, we give users the opportunity to configure defaults on `Faker<T>`'s `Faker` instance before it is used. Additionally, allow static `Date.DefaultDateTime` and `Date.DefaultDateTimeOffset` to be set. I don't know.

This is a difficult situation. This is one reason why I haven't really moved on #139 yet. I don't know which is the right answer without thinking more about the design and problem. Normally, I'd prototype several ideas to really get a feel for the right API design. I think that's what I'd have to do to resolve this issue.

### Re #139 
Another reason why I haven't moved on #139 yet is it's relatively easy to transform any `DateTime` or `DateTimeOffset` into UTC by calling `.ToUniversalTime()`. It's an ugly workaround but works and can be shorted with an extension method.

Also, there were [some enhancements](https://github.com/bchavez/Bogus/commit/dcfda64d7eb19a58fbdea48a6091761be351deb3#diff-f396bceb2fb63a74d5716b1d2405070fR76) I made to the `Date` dataset shortly after your #160 PR, which added a `refDate` to `.Soon(...)` and `.Recent(...)` that can take in a reference date for pinning time calculations to a specific reference point and respects `DateTimeKind`. For example:

```csharp
var f = new Faker();
var soonResult = f.Date.Soon( refDate: DateTime.UtcNow );
soonResult.Kind.Should().Be(DateTimeKind.Utc); //True
```

So, for now, as it stands, I don't feel this would be a good merge for the reasons I've outlined above. These are my thoughts. Let me know what you think. I hope they make some sense.

Thanks,
Brian

:hourglass_flowing_sand: :mag: ***["But I still haven't found what I'm for..."](https://www.youtube.com/watch?v=wdCJRybAtso)***

SimonCropp:
ok how about in the short term we drop the faker changes from this PR and run with only the static date.cs changes?

bchavez:
Yeah, I think dropping the `Faker` changes in the PR is a step in the right direction. The `Func<T>` constructor parameters are too loud in `Faker` and the `Date` dataset config details are bleeding up.

Below is a random, ***unfiltered*** raw dump of ideas that are currently floating in my head on the subject:

----

The **static** `DateTime` configuration seem okay. The drama is still seems around instance configurations.

* What if you just set one `DateTime`, given any `DateTime`, we can get it's UTC value. Managing one value might be better than managing two. What are the possible implications of doing this?
* Mulling over the constructor changes in `Date` too; kinda leaning that it's loud. My feels on `new Date("en", () => DateTime.UtcNow, ...)` mmm again, has noise, but I can understand it's use for setting the default time on an instance `Date` dataset. Still, something nagging inside me that it doesn't feel quite right. Maybe I need to just get over it. Haha
* I don't want to pollute the `.RuleFor( f => f.Date.Xyz())` API with configuration instance methods that could potentially confuse users. `.RuleFor( f => f.Date.ConfigureNow( () => ... ))`. Nah. Need to look for a better way to differentiate dataset configuration and regular use.
* What if `Date` instance `Func<DateTime> defaultNow` now was `protected internal` and could only be modified by including a configuration namespace (ie: `Bogus.Configuration`) that lights up extension method configuraiton APIs that sets these "default" dataset instance values. So,`faker.Date.ConfigureDefaultNow(() => DateTime.Now, ...))` can be called to set the dataset's internal "now" time configuration. Dunno, might be a lot of ceremony; discoverability not so good.
* What if `Date` instance had `faker.Date.DefaultDateTime` exposed as a public field. It does differentiate at least between method use and field use. Might be okay, huge OOP data encapsulation violation, but who's judging? Or another property? eh...
* Keep thinking...

:wrench: :hammer_and_wrench: ***["I got to modify... gotta multiply velocity..."](https://www.youtube.com/watch?v=I1NV97UYtCQ)***

SimonCropp:
constructor params removed

bchavez:
Hi Simon,

Thank you for reverting the changes. I appreciate it.

I'll have to take a closer look at this a little later. I'm a little preoccupied with a contract job at the moment. I should be back in full swing in about 15 days from now. Also, 15 days should give me some time to allow this to bake in the back of my mind for a little bit more. :smiley: 

Thanks,
Brian

:ear_of_rice: :stew: ***["Out here in the fields... I fight for my meals... It's only teenage wasteland..."](https://www.youtube.com/watch?v=IXWNSb4nUDY)***

bchavez:
Hi Simon,

I'm going move on this soon.

After thinking about it, I'm planning on only allowing only `DateTime` to be set from `Date`'s dataset. From some testing I've done, `DateTimeOffset` can be calculated or derived from the set `DateTime` with relative ease (but not so much from the reverse direction). For example:

linqpad_1966

Only allowing one parameter to be set avoids the possible misuse of setting two parameters that should represent the same thing, an instance of time.

The reason I'm picking `DateTime` as the data type to be set is it seems to be the lowest common denominator & most problematic type. `DateTime` has an association with `DateTimeKind` which seems difficult maintain going from `DateTimeOffset` -> `DateTime`.

If you have any thoughts on the subject or any foresight into the implications/limitations that this design choice could cause, let me know.

Thanks,
Brian

Issue #89, opened 2017-08-17
Error in designer when used in constructor


JerryNixon:
This works at runtime, this is an issue of during design time in the UWP-XAML designer. This occurs when the constructor of the view-model has the `Generate()` code. Listen, I really can't figure out the cause at this point; but, am very interested to integrate this library in the Template10.SampleData library and start teaching on Microsoft Virtual Academy using it. Thanks for looking.

> Could not load file or assembly 'Bogus, Version=15.0.6.0, Culture=neutral, PublicKeyToken=fa1bb3f3f218129a' or one of its dependencies. The system cannot find the file specified.

![image](https://user-images.githubusercontent.com/1749983/29395895-109f33ba-82c9-11e7-86dc-df60772d1fd3.png)

bchavez:
Hey @JerryNixon ,

Could you try installing [`System.Runtime`](https://www.nuget.org/packages/System.Runtime/). I'm not totally sure what's going on but this is the second time this has happened. Previously, another user `vbehara` encountered the same issue and posted in the [gitter channel](https://gitter.im/bchavez/Bogus).

We definitely need to figure out what's going on here.

:boom: :fire: ***["Set it ablaze like a candle wick... Light it up, light it up..."](https://www.youtube.com/watch?v=qDcFryDXQ7U)***

bchavez:
Also, if you could send me a zip sample of your project I'll be happy to kill this bug with fire :beetle: :fire:.

JerryNixon:
Well, with the Universal Windows Platform, there really isn't a `system.runtime`. Is there?

bchavez:
I found it odd too but `vbehara` mentioned it as his solution. Might try:
```
Install-Package System.Runtime -Version 4.3.0
```
This isn't really practical and **Bogus** should work out of the box with UWP.

I suspect there is some edge case I'm hitting (or I'm doing something really dumb) with NuGet tooling that pops up under UWP projects. I'm guessing NuGet and UWP are loading up the .NET Standard version of the Bogus DLL and it's got something to do with .NET Standard.

I don't have any UWP stuff installed on my local dev box; so this might take me a little longer to debug.

JerryNixon:
It installed, yeah. But it didn't correct the err. The proj I am working on can be zipped here, I'll do it to help you resolve this. I am sure it's simple. I will be presenting this at VSLive tomorrow and will skip the design-time data part for now. It's fine.

JerryNixon:
[...]

PR #166, opened 2018-07-22
add countrydata to readme

SimonCropp:


bchavez:
As always, great work Simon!

Thanks again for removing **Newtonsoft** dependency. Some **Bogus** users have had some issues with **Newtonsoft** in the past. Re: #98 and #89.

Thanks,
Brian

PR #194, opened 2019-01-17
Added first names to NL Locale

bartdebever:
To start off, I've seen that it is recommended to send in a pull request to the Faker.js repository but as that hasn't been updated in months and has 77 pull requests open, I fear it will not matter.

I found the company name generation for the Dutch locale to be lacking. I've ran into creating duplicate names in a system that requires them to be unique.

For this I've added a list of names gathered from [Wikipedia](https://nl.wikipedia.org/wiki/Lijst_van_voornamen), this doesn't have an English version.
I've also added more suffixes for companies and street suffixes.

I've ran the unit tests and hope the .bson file works, I don't have experience with this file format and done this sort of blind.


bartdebever:
Order has been changed but doesn't seem to effect unit tests.

bchavez:
Hi Bart,

No problem. Thank you for the PR. I really appreciate it.

 Actually, any modifications to the JSON files inside `data` will get clobbered on the next `gulp import.locales`. The best way to add extra data to a locale is to use the `data_extend` folder. JSON files in `data_extend` get merged from upstream faker.js then land in the `data` folder. This way we can constantly stay up to date with upstream faker.js while still making improvements to the JSON data.

For the JSON arrays that are "partially" updated, you'd have to copy the full JSON key to `data_extend`.

I can make these changes, or you can update the PR, whichever, just let me know. :+1: 

Thanks again, I really appreciate it. Once the changes are made, I'll make a new release!

Thanks,
Brian

Source/Bogus/data/nl.locale.json
```diff
@@ -3142,56 +3165,1590 @@
...
-      "Sem",
-      "Sophie",
-      "Stijn",
-      "Sven",
-      "Thijs",
-      "Thijs",
-      "Thomas",
-      "Tim",
-      "Tom"
+      [
```
bchavez: This is actually a bug here. Two nested JSON arrays `[ [`, but I'll fix it.

bchavez:
New version has been released. Thanks for your contribution! :sunglasses: 

https://www.nuget.org/packages/Bogus/25.0.4

PR #196, opened 2019-01-18
Replaced string with stringbuilder and code quality for finance

bartdebever:
- Made the StringBuilder changes not have the previous commits
- Upgraded the code quality of the Finance class

Code Quality is mostly because of a combination of the CodeMaid and Roslyn analysers, it's both a blessing and a sin as it does add bulk to the PR.

Hoping this change is small enough.

### DIFF
--- a/Source/Bogus/DataSets/Finance.cs
+++ b/Source/Bogus/DataSets/Finance.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 using Bogus.Bson;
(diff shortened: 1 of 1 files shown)


bchavez:
Great work Bart!

Thank you! :sunglasses: 

Let me know how urgent these changes are that you need. I can make another release, but if you have more incoming I'll wait.

-Brian

bartdebever:
As it's just a general performance improvement, they aren't urgent at all.?
Could just be an addition to a future release.

bchavez:
Okay cool. :+1: I was just thinking since you noticed the performance issue in your app that maybe it was causing some issues for you.

PR #197, opened 2019-01-20
DataSet class documentation

bartdebever:
- Added some documentation to the DataSet class.
- Changed some code to have high code quality.

Personally using [StyleCop](https://github.com/StyleCop/StyleCop)'s rules on how documentation should be done. I'd love to clean this class up more but I simply don't know enough about it.

### DIFF
--- a/Source/Bogus/DataSet.cs
+++ b/Source/Bogus/DataSet.cs
@@ -48,7 +48,7 @@ namespace Bogus
       private Randomizer randomizer;
 
       /// <summary>
-      /// Gets or sets the random used to generate values.
+      /// Gets or sets the <see cref="Randomizer"/> used to generate values.
       /// </summary>
       public Randomizer Random
       {
@@ -60,6 +60,11 @@ namespace Bogus
          }
       }
 
+      SeedNotifier IHasRandomizer.GetNotifier()
+      {
+         return this.Notifier;
+      }
+
       /// <summary>
       /// Resolves the 'category' type of a dataset type; respects the 'DataCategory' attribute.
       /// </summary>
@@ -69,11 +74,6 @@ namespace Bogus
       {
          var categoryAttribute = type.GetCustomAttributeX<DataCategoryAttribute>();
          return categoryAttribute != null ? categoryAttribute.Name : type.Name.ToLowerInvariant();
-      }
-
-      SeedNotifier IHasRandomizer.GetNotifier()
-      {
-         return this.Notifier;
       }
 
       /// <summary>
@@ -188,7 +188,7 @@ namespace Bogus
       /// <returns>The parsed token.</returns>
       private string ParseTokens(string value)
       {
-         return parseTokensRegex.Replace(
+         var parseResult = parseTokensRegex.Replace(
             value,
             x =>
                {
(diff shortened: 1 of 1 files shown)


Source/Bogus/DataSet.cs
```diff
@@ -11,68 +11,76 @@ namespace Bogus
...
       private Randomizer randomizer;
 
       /// <summary>
-      /// The Randomizer
+      /// Gets or sets the random used to generate values.
       /// </summary>
       public Randomizer Random
       {
-         get => this.randomizer ?? (this.Random = new Randomizer());
+         get => this.randomizer ?? (this.randomizer = new Randomizer());
```
bchavez: Hey Bart,

I think only setting `this.randomizer` variable is a breaking change.

Here's why: I think this should be reverted back to `this.Random` because `this.Random` causes the set method to execute 1) setting `this.randomzier` and 2) firing `this.Notifier.Notify(value)` to notify all dependency datasets that the parent randomizer has changed.

So, this change essentially breaks the notification mechanism which is critical for deterministic behavior. You can read more about these internals here: https://github.com/bchavez/Bogus/pull/101

Source/Bogus/DataSet.cs
```diff
@@ -91,18 +100,18 @@ protected internal virtual BValue Get(string category, string path)
...
       /// </summary>
+      /// <returns>A boolean to indicate if the locale exists.</returns>
       protected internal virtual bool HasKey(string path, bool includeFallback = true)
       {
-         if( includeFallback )
-            return Database.HasKey(this.Category, path, this.Locale);
-
-         return Database.HasKey(this.Category, path, this.Locale, null);
+         return includeFallback ? Database.HasKey(this.Category, path, this.Locale)
+                   : Database.HasKey(this.Category, path, this.Locale, null);
```
bchavez: It would be cool if this one could be reverted back to the original `if` statement. I'm not particularly a fan of log ternary operator lines that span multiple lines like this. At most, I'm okay if you wrap the original `if` bock with `{ }`.

In particular, I prefer to have the code in more complicated classes like this `DataSet` class to be slightly more on the verbose side as it helps with debugging issues since this is one of the core classes in **Bogus**.

Source/Bogus/DataSet.cs
bartdebever: Yup that was my mistake and stupid if I'm honest. Reverted the change.

Source/Bogus/DataSet.cs
bartdebever: ~~For the code Quality!~~ In all seriousness I was skeptical about this changes myself because it wasn't *that* readable so I reverted it, my apologies.

bchavez:
Thanks again Bart!

Yep, as a guiding principal, we want to maximize source-code readability, maintainability, and debug-ability as vectors for maximum success.

So, if you ever feel that some change did not fit into those three categories of awesomeness, then it might be good to revert it back. :sunglasses: 

:walking_man: :walking_woman: **[Missing Persons - Walking in L.A.](https://www.youtube.com/watch?v=R_UpLtGEWoY)**

PR #198, opened 2019-01-21
Rewrote doc of Address class.

bartdebever:
- Rewrote the documentation
- Added some blank lines
- Changed formatting of if-statements

Nothing much this time around.

### DIFF
--- a/Source/Bogus/DataSets/Address.cs
+++ b/Source/Bogus/DataSets/Address.cs
@@ -147,12 +147,12 @@ namespace Bogus.DataSets
       /// <returns>A random country code.</returns>
       public string CountryCode(Iso3166Format format = Iso3166Format.Alpha2)
       {
-         if (format == Iso3166Format.Alpha2)
+         if( format == Iso3166Format.Alpha2 )
          {
             return GetRandomArrayItem("country_code");
          }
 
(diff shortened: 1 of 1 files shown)


bchavez:
Hey Bart,

Looking good!

One thing, is it possible for you to change your project-level setting for the space in the `if` style block? For example, my IDE settings are currently configured this way:
```
if( expression )
{
}
```
Whereas yours is formatting the code like this:
```
if (expression)
{
}
```
It's basically bike-shedding at this point, but if we can get this style setup right, that would finally resolve or IDEs from fighting each-other and churning white space on every file re-format. :)

bartdebever:
> Hey Bart,
> 
> Looking good!
> 
> One thing, is it possible for you to change your project-level setting for the space in the `if` style block? For example, my IDE settings are currently configured this way:
> 
> ```
> if( expression )
> {
> }
> ```
> Whereas yours is formatting the code like this:
> 
> ```
> if (expression)
> {
> }
> ```
> It's basically bike-shedding at this point, but if we can get this style setup right, that would finally resolve our IDEs from fighting each other and churning white space on every file re-format. :)

Would need to dig deep into some settings from multiple extensions and doubt I can get my warnings away. I'll fight the compiler and try to keep them that way.

PR #216, opened 2019-04-03
Match heading to API usage ex: Faker.Phone.Phonenumber

marky291:
This is the only heading that does not match the API type-hinting, I feel this would benefit future users who were also confused by not being able to find phonenumbers.


bchavez:
Thank you :)

PR #237, opened 2019-07-14
Generate FR security social numbers

MarouanRejeb:


bchavez:
Hi @MarouanRejeb, before I release, can you point me to a Wikipedia article that describes France NSS numbers?

Thanks,
Brian

bchavez:
The closest I could find is an INSEE code here: https://en.wikipedia.org/wiki/INSEE_code

bchavez:
Actually, I'm really sorry @MarouanRejeb. I'm going to revert this PR/commit because it looks like there are checksums and modulo arithmetic that's required to pass validation of INSEE numbers.

If we really want to add FR social security numbers in Bogus, we'll have to follow the general format and algorithm more closely outlined here:

> Their format is as follows: syymmlloookkk cc,[1] where
> 
> * s is 1 for a male, 2 for a female,
> * yy are the last two digits of the year of birth,
> * mm is the month of birth, usually 01 to 12 (but there are special values for persons whose exact date of birth is not known),
> * ll is the number of the départment of origin : 2 digits, or 1 digit and 1 letter in metropolitan France, 3 digits for overseas.
> * ooo is the commune of origin (a department is composed of various communes) : 3 digits in metropolitan France or 2 digits for overseas.
> * kkk is an order number to distinguish people being born at the same place in the same year and month. This number is the one given by the Acte de naissance, > * an official paper which officialize a birth (and is needed throughout life for various administrative procedures, such as getting an identity card).
> * 'cc' is the "control key", 01 to 97, equal to 97-(the rest of the number modulo 97) or to 97 if the number is a multiple of 97.
>
> There are exceptions for people in particular situations.[1]

PR #245, opened 2019-08-06
Cleanup of the Lorem DataSet class.

bartdebever:
Cleanup of the documentation and variable names within the Lorem DataSet class.

FYI I started work on an [extension](https://github.com/bartdebever/VideoGames.Bogus) to Bogus myself for personal use. If this goes well I'd be down to make more of these extensions might I need them in the future.

### DIFF
--- a/Source/Bogus/DataSets/Lorem.cs
+++ b/Source/Bogus/DataSets/Lorem.cs
@@ -8,9 +8,11 @@ namespace Bogus.DataSets
    /// </summary>
    public class Lorem : DataSet
    {
+
       /// <summary>
-      /// Default constructor
+      /// Initializes a new instance of the <see cref="Lorem"/> class.
       /// </summary>
+      /// <param name="locale">The locale to be used to generate random values.</param>
       public Lorem(string locale = "en") : base(locale)
       {
       }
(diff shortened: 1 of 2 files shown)


Source/Bogus/DataSets/Lorem.cs
```diff
@@ -24,53 +26,61 @@ public string Word()
...
       /// Get a random sentence of specific number of words. 
       /// </summary>
-      /// <param name="wordCount">Get a sentence with wordCount words. Defaults between 3 and 10</param>
-      /// <param name="range">Add anywhere between 0 to 'range' additional words to wordCount. Default is 0.</param>
+      /// <param name="wordCount">
+      /// Get a sentence with wordCount words. Defaults between 3 and 10
+      /// </param>
+      /// <param name="range">
+      /// Add anywhere between 0 to 'range' additional words to wordCount. Default is 0.
+      /// </param>
```
bchavez: Hi Bart. Thanks for the PR. I just have a few stylistic issues here. I prefer to have these XML docs `<param>`s to be as compact as possible. IE:

Preferred: `<param>fff</param>` 
vs.
```
<param>
fff
</param>
```

Mostly because:
* Sometimes I have parsers that dig into these XML docs and having to deal with new line terminated text between `<tag>`s can be an extra annoyance.
* Also, more compact XML the better, this way most of the screen real estate is taken up by code, no so much multi-line XML docs.

I'm okay with `<summary>` being multi-line, but no so much `<param>`.

Please, don't take it personally, I'm mostly nitpicking at this point; but if we're looking forward to other data sets I'd like to mention it before we get too far.

I always appreciate your PRs.

Source/Bogus/DataSets/Lorem.cs
```diff
@@ -24,53 +26,61 @@ public string Word()
...
       public string Sentence(int? wordCount = null, int? range = 0)
       {
-         var wc = wordCount ?? this.Random.Number(3, 10);
+         wordCount = wordCount ?? this.Random.Number(3, 10);
```
bchavez: I'm not sure about overwriting the parameter values out-right here. Sometimes it's okay, sometimes not. Also, some of these methods are almost 1-1 with faker.js; which I'd prefer for maintenance reasons keep some of those variable names. In case there's ever an update to one of these methods, porting/updating these methods with same variable names helps ease the process. When we go about changing variable names like this it can introduce some friction for me to rebuild the program state in my head to get a picture of what variable holds what state and makes the maintenance/porting process more difficult. I don't know if this is one of those methods, but when possible, try to avoid renaming variables if the variables are the same used in faker.js. If there's is something more fundamental about the algorithm that can be more idiomatic in C#, then we can discuss changing it.

Source/Bogus/DataSets/Lorem.cs
```diff
@@ -24,53 +26,61 @@ public string Word()
...
       /// Get some sentences.
       /// </summary>
       /// <param name="sentenceCount">The number of sentences</param>
+      /// <param name="separator">The string to separate the sentences</param>
       public string Sentences(int? sentenceCount = null, string separator = "\n")
       {
-         var sc = sentenceCount ?? this.Random.Number(2, 6);
-         var sentences = Enumerable.Range(1, sc)
-            .Select(s => Sentence());
+         sentenceCount = sentenceCount ?? this.Random.Number(2, 6);
```
bchavez: same as above comment.

Source/Bogus/DataSets/Lorem.cs
```diff
@@ -123,16 +139,23 @@ public string Text()
...
+      /// <param name="lineCount">
+      /// The amount of lines to be generated. Defaults between 1 and 5.
+      /// </param>
+      /// <param name="separator">
+      /// The string to separate the lines
+      /// </param>
       public string Lines(int? lineCount = null, string separator = "\n")
       {
-         var lc = lineCount ?? this.Random.Number(1, 5);
+         lineCount = lineCount ?? this.Random.Number(1, 5);
```
bchavez: same as above comment.

Source/Bogus/DataSets/Lorem.cs
bartdebever: Understandable, didn't know this was more of a direct port. Reverted these instances.

Source/Bogus/DataSets/Lorem.cs
bartdebever: It's always good to mention these kinds of things, I'll try to keep it in mind for future PRs. Don't worry about nitpicking, these PRs are already mostly my documentation OCD kicking in. You are the owner of the repository so I'll go by your rules.

bchavez:
Thanks so much for making the changes Bart.

I will finish your PR merge tonight or tomorrow. Right now, I have some things I need to attend to. Again, I appreciate your work and help very much. :smiley: I apologize for the delay.

PR #254, opened 2019-09-08
Fixes #253 Adds Support for RuleFor(string) and Ignore(string)

### LINKED ISSUE #253: String overloads for .RuleFor and .Ignore
Hey @bchavez,

I'm bringing the discussion from [**AutoBogus**](https://github.com/nickdodd79/AutoBogus/issues/21) over here, so we can consider the changes.

For anyone wanting to catch up, follow the [link](https://github.com/nickdodd79/AutoBogus/issues/21) to read @bchavez considerations, and my own, so far

**Context**

Currently, there is no simple way to configure a rule for hidden members other than inherit from `Faker`.

**Alternatives**

As above says, to accomplish this, we have to inherit from `Faker` and use already existing protected members.

**Has the feature been requested before?**

Not really, but it was requested over in [**AutoBogus**](https://github.com/nickdodd79/AutoBogus/issues/21)

**If the feature request is approved, would you be willing to submit a PR?**

I believe that exposing these members is more a design than technical concern, but given enough time (something we all need...), I'd be happy to help.

bchavez:
Adds support for Faker<T>.RuleFor("prop", ...) and Faker<T>.Ignore("prop") where "prop" is a property or field name on `T`. This is helpful in situations where it might be necessary to set up a rule for non-public members; such as `protected` members.

Mostly renamed the existing `protected virtual RuleFor(string)` to `protected virutal RuleForInternal(string)`.

Then added `public virtual RuleFor(string)` which forwards calls to `RuleForInternal`; but before it does so, ensures that the underlying member exists (from the point of view of the `IBinder`).

After looking at the existing code; I found `.Ignore` checks the existence of the members on `T` at `.Ignore` invocation time.  Considering that we already do this for `.Ignore`, I added the same precondition constraint to `.RuleFor(string)` as well.

Basically, the member must exist when calling `.RuleFor(string)` or `.Ignore(string)` for consistency. Plus, it sets up a strong condition at development time to ensure developers are writing valid rules; instead of silently failing at runtime due to no type-safety through string.

//cc @SuricateCan @nickdodd79

### DIFF
--- a/Source/Bogus.Tests/GitHubIssues/Issue134.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue134.cs
@@ -49,7 +49,7 @@ namespace Bogus.Tests.GitHubIssues
             var memberName = parts[0].Trim();
             var handlebarRule = parts[1].Trim();
 
-            this.RuleFor(memberName, f => f.Parse($"{handlebarRule}"));
+            this.RuleForInternal(memberName, f => f.Parse($"{handlebarRule}"));
          }
 
          return this;
--- a/Source/Bogus.Tests/GitHubIssues/Issue253.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue253.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using Bogus.Tests.Models;
+using FluentAssertions;
+using Xunit;
+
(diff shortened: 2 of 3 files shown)


SuricateCan:
Great work @bchavez, looks good for me.

Issue #255, opened 2019-09-20
Properties on parent interfaces not populated


rowland-banks-abide:
### Version Information
| Software                       | Version(s) |
| ------------------------| ---------- |
| Bogus NuGet Package  |    28.2.1        |
| .NET Core?                   |      Y           |
| .NET Full Framework?   |      Y          |
| Windows OS?              |        Y         |
| Linux OS?                    |                  |
| Visual Studio?              |        2019      |

### What is the expected behavior?
Property should have a value

### What is the actual behavior?
Property is null.

### Any possible solutions?
Yes - I've fixed the issue and will be submitting a PR shortly.
### How do you reproduce the issue?
```
void Main()
{
	var child = new Faker<IChild>()
		.CustomInstantiator(f => new Child())
		.RuleFor(e => e.Name, f => f.Person.FullName)
		.Generate();
		
	if (child.Name == null)
	{
		Console.WriteLine("Oh no - we have a bug!");
	}
}

public interface IParent
{
	string Name { get; set; }
}

public interface IChild : IParent
{
	string City { get; set; }
}

public class Child : IChild
{
	public string Name { get; set; }
	public string City { get; set; }
}
```

### Do you have a unit test that can demonstrate the bug?
Yes.

### Can you identify the location in Bogus' source code where the problem exists?
Yes - I've fixed the issue and will be submitting a PR shortly. 

### If the bug is confirmed, would you be willing to submit a PR?
Yes

Yes / No _(Help can be provided if you need assistance submitting a PR)_

bchavez:
Hi Rowland,

Thank you for the report. This is a pretty interesting use case. I never thought about using `Faker<T>` with interfaces like this.

My first reaction is to recommend using `Faker<Child>` then cast the generated result to `IChild`.

I'll consider supporting this use case on `Faker<T>`, but first let's check out how your PR looks and we'll go from there. As long as it doesn't have a significant impact on maintainability, existing unit tests, and Bogus Premium's Analyzer; it might be okay.

Thanks,
Brian

:mount_fuji: :man_playing_water_polo: **[Creedence Clearwater Revival - Green River](https://www.youtube.com/watch?v=L5V9nK7-OkM)**

bchavez:
Hi Rowland,

Sorry for the delay, **Bogus `v28.3.1`** is now live with your changes here: https://www.nuget.org/packages/Bogus/28.3.1

Thanks again for the concise and clean PR; also, thanks for following the coding style. :+1:  
-Brian

PR #256, opened 2019-09-20
Fixes issue 255

rowland-banks-abide:
Issue was that members defined on parent interface are not retrieved via a simple `Type.GetMembers()` call.

I've added call to `GetMembers` for every interface in the inheritance tree.

Fixes https://github.com/bchavez/Bogus/issues/255.


bchavez:
Thank you Rowland. I appreciate your clean PR very much.

Please give me a few hours to review the situation. I'll get back to you soon.

rowland-banks-abide:
Thanks for merging it so promptly, and for making such an amazing library :)

PR #258, opened 2019-10-03
add Path to Internet dataset

AlexCuse:


AlexCuse:
Let me know if you want me to squash the two commits before you merge.  I didn't see the methods listed in readme until after I opened this.

bchavez:
Hi Alex,

Thank you for the pull request. I appreciate the effort you've put into the PR here; however, I do have some concerns.

I'm not convinced yet about adding a public `Internet.Path()` method. There is something about it API wise and semantically that doesn't fit right with me yet. When I read `f => f.Internet.Path()` am I getting a full-path? file-path? relative-path? or absolute path? It is not clear.

Also, do we need all this ceremony when the following would do just fine (either as-is or as an extension method):

```csharp
var f = new Faker(); 
Path.ChangeExtension(f.Internet.UrlWithPath(), ".txt").Dump();
```
```
https://dallas.org/sleek-concrete-tuna/initiatives.txt
```

I'll try to take a second look tomorrow, it's 12:02 AM here so, it might be good for me to get some rest and sleep on this a bit.

Thanks,
Brian

AlexCuse:
I don't really care much about the extension for my current use case, but could envision it being useful for some test scenarios.   Mostly just wanted to be able to easily get a valid [PathString](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.pathstring?view=aspnetcore-2.2) value for implicit conversion.  When using a normal random string it throws

`System.ArgumentException : The path in 'value' must start with '/'.
Parameter name: value
   at Microsoft.AspNetCore.Http.PathString..ctor(String value)`

 Named it Path just to match property names on the HTTP request I was mocking but RelativePath would work too.  Figured if it was next to Url and UrlWithPath it all kind of fits but theres nothing wrong with being explicit.  If you like that name better let me know, I can change it and squash the two commits together at the same time.

bchavez:
Hi Alex,

Thanks for your patience. Also, thank you for providing context for your usage of `.Path`.

I'm still not sure about the `.Path` method name. We kinda have something already similar to `.Path` on `System.FilePath()`; although, it generates more `/etc/`-linux like file system paths. This `.Path` method looks more like user-generated paths.

I kind of like the idea of tacking on `fileExt` parameter to `.UrlWithPath(...)` to generate file paths with a specific file extension.

It's just when I have indecision like this, typically what happens is I ask users to define their own "formatting things" into user-defined extension methods that hang off `Internet` that allows you to specifically format URLs to your exact scenario.

I still need to think about this a little more: whether something like `.Path` should belong in `System` instead or make `.Path` private, or change it to `.UrlRelativePath`; or something more fitting.

Again, I apologize for the indecision here; just that new public APIs go through a little more scrutiny w.r.t semantics and ergonomics.

Thanks,
Brian

AlexCuse:
Cool let me know.  FWIW I think Internet.RelativePath would make pretty good sense.  I do like the idea of an extension method for the extension stuff.  I will poke around and see if I can find any examples.

bchavez:
Hi Alex,

I'm so sorry for the delay. I released **Bogus** `v28.4.1`: https://www.nuget.org/packages/Bogus/28.4.1

The Bogus `Internet` dataset should now have `Internet.UrlRootedPath(fileExt)` and `Internet.UrlWithPath(fileExt)` options.

Let me know how it works for you. I hope that helps. I decided to use a slightly different approach than this PR, but it should work the same.

My apologies it took so long. I just started a new job and my OSS work has been delayed a little bit.

Thanks,
Brian

AlexCuse:
This is great thanks!

PR #259, opened 2019-10-04
Fix GenerateForever to pass rulesets parameter

StanleyGoldman:
I was trying to use `GenerateForever` and I noticed that the `ruleSets` parameter is not utilized.


bchavez:
Hi @StanleyGoldman 

Thanks very much for the PR.

Your changes are now live in **Bogus** `v28.3.2`: https://www.nuget.org/packages/Bogus/28.3.2

Let me know if you have any issues.

Thanks again!
Brian Chavez

PR #265, opened 2019-11-05
Fix wrong namespace for Brazilian CNPJ Extension

EduTedeschi:


bchavez:
Hi @EduTedeschi,

Thank you very much for the correction. I really appreciate it. I have to leave for work right now, but we'll look at getting this merged in later tonight.

Thanks again for your PR!
-Brian

bchavez:
Hi @EduTedeschi,

After looking at others in the README and looking at the code, I found:

* If we changed it to `Bogus.Company.Cnpj()` the namespace and path would be incorrect.
* As it is now, the `Bogus.DataSets.Company.Cnpj()` would be the correct namespace and path because of the following:
    ```csharp
  var c = new Bogus.DataSets.Company();
  var cnpj = c.Cnpj();
    ```

Thanks,
Brian

PR #269, opened 2019-11-29
Add includeFormatSymbols for cnpj & cpf in brazil extensions

ArthNRick:
Often developers need this data (cpf and cnpj) to be numeric, unformatted, that's what this implementation does.

### DIFF
--- a/Source/Bogus.Tests/DataSetTests/CompanyTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/CompanyTest.cs
@@ -54,9 +54,15 @@ namespace Bogus.Tests.DataSetTests
       }
 
       [Fact]
+      public void can_generate_cnpj_for_brazil_without_formatting()
+      {
+         company.Cnpj(includeFormatSymbols: false).Should().Be("61860606000191");
+      }
+
+      [Fact]
       public void can_generate_numeric_cnpj_for_brazil()
       {
-         company.NumericCnpj().Should().Be((ulong)61860606000191);
+         company.CnpjNumeric().Should().Be(61860606000191);
       }
 
       [Fact]
--- a/Source/Bogus.Tests/DataSetTests/CompanyTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/CompanyTest.cs
@@ -60,12 +60,6 @@ namespace Bogus.Tests.DataSetTests
       }
 
       [Fact]
-      public void can_generate_numeric_cnpj_for_brazil()
-      {
-         company.CnpjNumeric().Should().Be(61860606000191);
-      }
-
-      [Fact]
       public void can_generate_an_EIN()
       {
          company.Ein().Should().Be("61-8606064");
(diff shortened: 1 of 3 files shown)


bchavez:
Hi Arthur,

Thank you for the Pull Request.

I refactored the code to include an `includeFormatSymbols` parameter in `Cnpj()` and `Cpf()`. When `includeFormatSymbols = false` the methods don't return formatted symbols.

For example,  `company.Cnpj(includeFormatSymbols: false)` returns `"61860606000191"` (string).

The reasoning to keep it as a string is mostly because as a `string`, it can be converted to `long` or `ulong` or be kept as an unformatted `string`.

In the initial PR, `NumericCpf()` returned `ulong` which is somewhat more restrictive to consumers who might want `long` or unformatted `string`.

This means however if you want to consume `Cnpj(includeFormatSymbols: false)` or `Cpf(includeFormatSymbols: false)` method, you'll need to `ulong.Parse()` for your own purposes.

**Bogus v28.4.2** is now available with these changes:  
https://www.nuget.org/packages/Bogus/28.4.2

```csharp
void Main()
{
   var c = new Company();
   
   var cnpjList = Enumerable.Range(1,10)
                  .Select(_ => c.CnpjNumeric())
                  .ToArray();
   cnpjList.Dump();
}

public static class MyCustomExtensionsForBogus
{
   public static ulong CnpjNumeric(this Company c){
      return ulong.Parse(c.Cnpj(includeFormatSymbols: false));
   }
}
```
LINQPad_3614


I hope that helps.

Thanks,
Brian

ArthNRick:
OK. were good adjustments, thank you very much!

ArthNRick:
The only problem with this solution is the context, when a value is requested with formatting, it will be impossible to get the same value without formatting, and vice versa. In separate methods, the context would eventually return the same value, and the method would replace.

bchavez:
Is Cpf the only one with Person context? Did context work in previous version?

ArthNRick:
> Is Cpf the only one with Person context? Did context work in previous version?

The CPF is for the Brazilian the same as the SSN is for the Americans, so the same person can not have more than one number, it is unique for each individual. Therefore it must be kept in context to prevent data from being randomized. The new version still works, but once I generate formatted data, I can no longer get it without formatting, since it is in context with the first value

bchavez:
Hi Arthur,

Thank you for the rely. I think what I'm trying to find out is if the
current CPF for Person with no context is the same as the previous version.
Does the previous version of Bogus before this PR have the same behavior of
no context when called on Person?

Ie:
```csharp
Var p = new Person()
p.Cpf()
p.Cpf()
```
Is the behavior the same in previous version of Bogus? If I recall
correctly, previous version has no context either. If that's the case then
we'll need a separate PR or issue. I'll try to investigate more later when
I'm off work.

ArthNRick:
Both the current and previous versions returned the same value by exiting the following code:
```cs
var p = new Person ();
var isEquals = p.Cpf () == p.Cpf () // true;
```

but with the latest version, `p.Cpf () == p.Cpf (false);` would return true as well, but should be false since p.Cpf (false) would not be formatting, and p.Cpf () formatted. I will create a new PR with a possible solution, and then you can evaluate if it is ideal

bchavez:
Ah, thank you @ArthNRick. I understand the issue now. We'll get that fixed.

PR #275, opened 2019-12-28
C# 8 and code quality improvements

SimonCropp:


Source/Benchmark/BenchSsn.cs
```diff
@@ -37,7 +36,7 @@ public void SsnAlgo2()
          var b = r.Int(1, 99);
          var c = r.Int(1, 9999);
 
-         var result = string.Format("{0:000}-{1:00}-{2:0000}", a, b, c);
```
bchavez: Hey Simon, actually, the use of `string.Format` was intentional here to compare any performance difference between interpolated C# strings and `string.Format`. Otherwise, the benchmark between `SsnAlgo1` and `SsnAlgo2` would be the same. `SsnAlgo2` was part of exploring the topic of formatting strings.

I'll help update the PR later after some more review today and tomorrow. I have to visit the store in a bit.

bchavez:
Thanks a bunch Simon! Fantastic work as always! :+1: :sunglasses: 

:chocolate_bar: :cookie: :lollipop: [**Ronald Jenkees - Stay Crunchy**](https://www.youtube.com/watch?v=GeIAXlwVlZc)

PR #277, opened 2020-01-02
WIP: Expose new overload for IRuleSet.RuleFor method

thyago-akira:
This commit exposes a new overload for IRuleSet.RuleFor method and fixes the tests with ambiguity errors


bchavez:
Hi Thyago,
You might want to hold off on making more changes. I only briefly reviewed some of the changes and I may not accept the PR because we are changing the public API.

I'll need to do a more thorough review of the when I get home, but I wanted to give you a quick heads up before investing too much time. 

Thanks, Brian

thyago-akira:
> Hi Thyago,
> You might want to hold off on making more changes. I only briefly reviewed some of the changes and I may not accept the PR because we are changing the public API.
> 
> I'll need to do a more thorough review of the when I get home, but I wanted to give you a quick heads up before investing too much time.
> 
> Thanks, Brian

Hi, Brian. 

No problem. The main change is already finished (it just exposes a new overload to the method), with some existing tests adjusted. 
I can wait until you do your review, but the only thing that's missing is a specific unit test for this PR (something similar to what I have already done in the other tests).

Thanks for your feedback,

Thyago Akira

bchavez:
Hi Thyago,

Thank you very much for your message.

I was wondering, could you explain the impetus for adding the new overload? What is the main reason for requesting the change in the source code? It would be very helpful if you could explain why you need this change. Also, if you could please provide a few examples of the limitations of the current API and contrast it to your proposed changes, that would help me very much.

I hope I don't sound rash, but usually, with changes like this you'd want to create a GH issue first to discuss any changes to the code, especially regarding a public API change, then get the okay from the maintainers before investing time. Nothing hurts more than having a PR rejected that a lot of time was spent on. 

**Also, quick tip:** only change files that are pertinent to the topic at hand. Avoid changing too many files (like code cleanups) that aren't related to your implementation because it makes it harder to review changes. 

I just need to fully understand where you're coming from and weight that against changes to the public API for potentially millions of downstream consumers.

Thanks,
Brian

Source/Bogus/IRuleSet.cs
```diff
@@ -23,6 +23,11 @@ public interface IRuleSet<T> where T : class
       /// </summary>
       Faker<T> RuleFor<TProperty>(Expression<Func<T, TProperty>> property, Func<Faker, TProperty> setter);
 
+      /// <summary>
+      /// Creates a rule for a property and providing access to the instance being generated.
+      /// </summary>
+      Faker<T> RuleFor<TProperty>(Expression<Func<T, TProperty>> property, Func<T, TProperty> setter);
+
```
bchavez: My current observation of this change is that adding this method to `IRuleSet<T>` causes a slue of source-code breaking changes, not just for us, but for others too. Most of the breaking changes seem to be "ambiguous" call compiler errors that the compiler is having a hard time resolving.

If your ultimate goal with these changes is to make the compiler automatically resolve the overload:
* `.RuleFor(o => o.OrderId, f => orders++);` // captured variable
* `.RuleFor(o => o.OrderId, f => f.Random.Int());` // faker facade
* `.RuleFor(o => o.OrderId, o => o.CreatedDate.Ticks);` // order object

You could probably use an extension method in your own project like the following example:

```csharp
public static class ExtensionsForFakerOfT
{
   /// <summary>
   /// Creates a rule for a property and providing access to the instance being generated.
   /// </summary>
   public static Faker<T> RuleFor<T, TProperty>(this Faker<T> faker, Expression<Func<T, TProperty>> property, Func<T, TProperty> setter)
      where T : class
   {
      return faker.RuleFor(property, (f, t) => setter(t));
   }
}
```
This workaround would be readily available to you and we wouldn't break existing source code in the wild.

thyago-akira:
> Hi Thyago,
> 
> Thank you very much for your message.
> 
> I was wondering, could you explain the impetus for adding the new overload? What is the main reason for requesting the change in the source code? It would be very helpful if you could explain why you need this change. Also, if you could please provide a few examples of the limitations of the current API and contrast it to your proposed changes, that would help me very much.
> 
> I hope I don't sound rash, but usually, with changes like this you'd want to create a GH issue first to discuss any changes to the code, especially regarding a public API change, then get the okay from the maintainers before investing time. Nothing hurts more than having a PR rejected that a lot of time was spent on.
> 
> **Also, quick tip:** only change files that are pertinent to the topic at hand. Avoid changing too many files (like code cleanups) that aren't related to your implementation because it makes it harder to review changes.
> 
> I just need to fully understand where you're coming from and weight that against changes to the public API for potentially millions of downstream consumers.
> 
> Thanks,
> Brian

Hi Brian,

Thanks for all your comments and your help. This isn't a requirement, but I was testing something on a fork, and it worked well. I understand that will break the public API, because the changes made me to fix some existing tests. That's why I still have this PR as draft, and I have no problem if you prefer to reject it,

And explaining the reason of that, I was trying to decouple part of my implementation of scenarios builder from the Bogus.Faker repository. If you want to take a look at it, here is the link:
https://github.com/thyago-akira/akira.testtools.scenarios

Regards,

Thyago Akira

bchavez:
Hi Thyago,

That's a really cool project! Pretty interesting. Keep up the great work! It looks very promising.

As far as changing Bogus' public API, I don't think I will be accepting this PR because the PR carries a high cost of potentially breaking everyone's code at compile time. I'm generally okay with breaking changes but in this case, I think the cost is too high.

This PR is an **improvement** to the public API, but the bar for breaking someone's code is very high. Additionally, the new API changes feel like there is an underlying force that requires retraining muscle memory when using the new API changes. The retraining force is not very big, but it is still there nonetheless.

Lastly, the workaround for your issue should be straight forward. The best part is you get the functionality you're looking for and would not break everyone's code. To decouple your API from `Faker` create an extension method (like you did in https://github.com/thyago-akira/akira.testtools.scenarios/pull/21) that would have the compiler resolve `T` instead of `Faker` as indicated in my review comment above:

```csharp
public static class ExtensionsForFakerOfT
{
   /// <summary>
   /// Creates a rule for a property and providing access to the instance being generated.
   /// </summary>
   public static Faker<T> RuleFor<T, TProperty>(this Faker<T> faker, Expression<Func<T, TProperty>> property, Func<T, TProperty> setter)
      where T : class
   {
      return faker.RuleFor(property, (f, t) => setter(t));
   }
}
```
Now, `.RuleFor(o => o.OrderId, o => o.CreatedDate.Ticks); // order object` should work.

I do appreciate your work and point of view regarding this PR. It is now something I wish I had implemented at the beginning; but given the current situation, it makes it difficult to accept this change.

I will keep a mental note of this API change and tag it as an enhancement for future improvements. If I get a high demand for a change like this I'll consider adding it to the public API in the future.

Thanks,
Brian Chavez

thyago-akira:
> Hi Thyago,
> 
> That's a really cool project! Pretty interesting. Keep up the great work! It looks very promising.
> 
> As far as changing Bogus' public API, I don't think I will be accepting this PR because the PR carries a high cost of potentially breaking everyone's code at compile time. I'm generally okay with breaking changes but in this case, I think the cost is too high.
> 
> This PR is an **improvement** to the public API, but the bar for breaking someone's code is also very high. Additionally, the new API changes feel like there is an underlying force that requires retraining muscle memory when using the new API changes. The retraining force is not very big, but it is still there nonetheless.
> 
> Lastly, the workaround for your issue should be straight forward. The best part is you get the functionality you're looking for and would not break everyone's code. To decouple your API from `Faker` create an extension method (like you did in [thyago-akira/akira.testtools.scenarios#21](https://github.com/thyago-akira/akira.testtools.scenarios/pull/21)) that would have the compiler resolve `T` instead of `Faker` as indicated in my review comment above:
> 
> ```cs
> public static class ExtensionsForFakerOfT
> {
>    /// <summary>
>    /// Creates a rule for a property and providing access to the instance being generated.
>    /// </summary>
>    public static Faker<T> RuleFor<T, TProperty>(this Faker<T> faker, Expression<Func<T, TProperty>> property, Func<T, TProperty> setter)
>       where T : class
>    {
>       return faker.RuleFor(property, (f, t) => setter(t));
>    }
> }
> ```
> 
> Now, `.RuleFor(o => o.OrderId, o => o.CreatedDate.Ticks); // order object` should work.
> 
> I do appreciate your work and point of view regarding this PR. It is now something I wish I had implemented at the beginning; but given the current situation, it makes it difficult to accept this change.
> 
> I will keep a mental note of this API change and tag it as an enhancement for future improvements. If I get a high demand for a change like this I'll consider adding it to the public API in the future.
> 
> Thanks,
> Brian Chavez

Hi Brian,

Thanks your feedback. I appreciate your time reviewing my project. If you have something that is missing there, please let me know. And I will continue following the Bogus project.

Regards,

Thyago Akira

PR #288, opened 2020-04-06
Update ru.locale.json

o4zloiroman:
Separated a city entity consisting of two different cities into two proper ones.


bchavez:
Hi @o4zloiroman,

Thank you for letting me know. This should be fixed in **Bogus** `v29.0.2`:
https://www.nuget.org/packages/Bogus/29.0.2

Thanks!
Brian

PR #300, opened 2020-06-11
Enhancements to Randomizer

logiclrd:
This PR proposes several enhancements to the implementation of Randomizer:

- It does away with the hard upper bound of `int.MaxValue - 1` in `Number`.
  * If the caller specifies a `max` value less than `int.MaxValue`, then it simply adds one to it to make it inclusive.
  * If the caller specifies `int.MaxValue`, then it can't add 1 to `max`, so instead it subtracts 1 from `min`, adding it back into the result.
  * If the caller also specifies `int.MinValue` for min, then it can't do that either, so it takes two samples, one of which is in the range `int.MinValue`..`-1`, and the other of which is used to decide 50% of the time to invert all the bits, producing a number the range `0`..`int.MaxValue`. (This determination is not based on the least significant bit, which I understand to be the least random from the stock `Random` class.) Thus, with even probability, it will be between 0x80000000 and 0xFFFFFFFF and 0x00000000 and 0x7FFFFFFF, which is the same as having a uniform distribution across 0x80000000 to 0x7FFFFFFF.

- It streamlines `Even` and `Odd` and updates them to ensure even distribution over the possible even or odd return values.
  * They adjust the input range so that the raw range contains the same number of even and odd numbers, always starting on an even number and ending on an odd number.
  * A sample is taken in this range, and then the least significant bit is set or cleared as needed to force the number to be odd or even. Since the sample set has the same number of even and odd values, and since we trust `Number` to be an even distribution (which it _almost_ is, it's twice as likely to pick the largest possible value than any other value because of a poorly-implemented clamp internally -- but this should be virtually unnoticeable), this should ensure that the result numbers have an equally even distribution.
  * The code specifically checks for boundary cases where the range adjustment math would break, and the behaviour for ranges that don't contain suitable values is clearly documented.

- It updates the way `decimal` values are generated. The old method, scaling a `double`, meant that many, many possible values were never possible. This is because a `double`'s mantissa is only 52 bits wide, whereas in `double` it is 96 bits wide. The lower 44 bits will effectively never contain any random data. The new generator exploits the (documented) structure of the bits in `decimal`, filling the full 96 bits of the mantissa with random data (exploiting the new `Number` which can satisfy the entire 32-bit range without excluding `int.MaxValue`), and then scaling and translating the value into the desired range in a way that avoids reducing the scale as much as possible.

- It updates the way `float` values are generated in a pretty trivial way, translating them into their desired range while they are still a `double` instead of truncating their precision first. This only really affects the least significant bit of the mantissa, as I understand it, but it is the principle of the matter. :-)

Unit tests are updated correspondingly. In particular, several tests assume specific sequences of values back from `.Int()`, and needed to have their expectations updated to match the changes this rework brings in. One test had a bug, where it would use `& 16383` to generate a 4-digit value. By chance, this generated the value `0111` before, but after these changes, a value larger than `9999` is produced. I added a line to the test to ensure these values are 4 digits long.

### DIFF
--- a/Source/Bogus.Tests/GitHubIssues/Issue124.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue124.cs
@@ -129,24 +129,34 @@ namespace Bogus.Tests.GitHubIssues
             .RuleFor(x => x.Obj, f => new object().OrNull(f))
             .RuleFor(x => x.Str, f => f.Random.Word().OrNull(f));
 
-         var q = faker.Generate(3);
+         var q = faker.Generate(5);
 
          console.Dump(q);
 
          q[0].Id.Should().NotBeNull();
          q[0].Gud.Should().NotBeNull();
-         q[0].Obj.Should().NotBeNull();
-         q[0].Str.Should().NotBeNull();
+         q[0].Obj.Should().BeNull();
+         q[0].Str.Should().BeNull();
 
-         q[1].Id.Should().NotBeNull();
+         q[1].Id.Should().BeNull();
          q[1].Gud.Should().NotBeNull();
-         q[1].Obj.Should().NotBeNull();
+         q[1].Obj.Should().BeNull();
          q[1].Str.Should().BeNull();
-         
+
          q[2].Id.Should().BeNull();
-         q[2].Gud.Should().BeNull();
+         q[2].Gud.Should().NotBeNull();
          q[2].Obj.Should().NotBeNull();
-         q[2].Str.Should().NotBeNull();
+         q[2].Str.Should().BeNull();
+
+         q[3].Id.Should().NotBeNull();
+         q[3].Gud.Should().NotBeNull();
+         q[3].Obj.Should().NotBeNull();
+         q[3].Str.Should().BeNull();
+
+         q[4].Id.Should().BeNull();
+         q[4].Gud.Should().NotBeNull();
+         q[4].Obj.Should().NotBeNull();
+         q[4].Str.Should().NotBeNull();
       }
 
       public class Foo
--- a/Source/Bogus.Tests/GitHubIssues/Issue124.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue124.cs
@@ -135,28 +135,28 @@ namespace Bogus.Tests.GitHubIssues
 
          q[0].Id.Should().NotBeNull();
          q[0].Gud.Should().NotBeNull();
-         q[0].Obj.Should().BeNull();
-         q[0].Str.Should().BeNull();
+         q[0].Obj.Should().NotBeNull();
+         q[0].Str.Should().NotBeNull();
 
-         q[1].Id.Should().BeNull();
+         q[1].Id.Should().NotBeNull();
          q[1].Gud.Should().NotBeNull();
-         q[1].Obj.Should().BeNull();
+         q[1].Obj.Should().NotBeNull();
          q[1].Str.Should().BeNull();
 
          q[2].Id.Should().BeNull();
-         q[2].Gud.Should().NotBeNull();
+         q[2].Gud.Should().BeNull();
          q[2].Obj.Should().NotBeNull();
-         q[2].Str.Should().BeNull();
+         q[2].Str.Should().NotBeNull();
 
-         q[3].Id.Should().NotBeNull();
+         q[3].Id.Should().BeNull();
          q[3].Gud.Should().NotBeNull();
-         q[3].Obj.Should().NotBeNull();
-         q[3].Str.Should().BeNull();
+         q[3].Obj.Should().BeNull();
+         q[3].Str.Should().NotBeNull();
 
          q[4].Id.Should().BeNull();
          q[4].Gud.Should().NotBeNull();
          q[4].Obj.Should().NotBeNull();
-         q[4].Str.Should().NotBeNull();
+         q[4].Str.Should().BeNull();
       }
 
       public class Foo
--- a/Source/Bogus.Tests/GitHubIssues/Issue260.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue260.cs
@@ -21,7 +21,7 @@ namespace Bogus.Tests.GitHubIssues
          var x = r.Int();
 
          // the remainder of this test depends on x having a predictable value
-         x.Should().Be(425714706);
+         x.Should().Be(1077349347);
 
          // right shift all bits except fir the first 10 bits = 2^10 = 1024.
          var a = (x >> (32 - 10)) % 898;
@@ -38,7 +38,7 @@ namespace Bogus.Tests.GitHubIssues
 
          var result = $"{a:000}-{b:00}-{c:0000}";
 
-         result.Should().Be("101-18-6328");
+         result.Should().Be("256-99-1799");
       }
    }
 }
--- a/Source/Bogus.Tests/GitHubIssues/Issue260.cs
+++ b/Source/Bogus.Tests/GitHubIssues/Issue260.cs
@@ -20,6 +20,9 @@ namespace Bogus.Tests.GitHubIssues
 
          var x = r.Int();
 
+         // the remainder of this test depends on x having a predictable value
+         x.Should().Be(425714706);
+
          // right shift all bits except fir the first 10 bits = 2^10 = 1024.
          var a = (x >> (32 - 10)) % 898;
          if( a == 0 || a == 666 ) a++;
@@ -30,11 +33,12 @@ namespace Bogus.Tests.GitHubIssues
 
          // last 2^14 = 16384, for last 4 digits of SSN
          var c = (x >> 7) & 0x3FFF;
+         if( c >= 10000 ) c -= 10000;
          if( c == 0 ) c++;
 
          var result = $"{a:000}-{b:00}-{c:0000}";
 
-         result.Should().Be("309-89-0111");
+         result.Should().Be("101-18-6328");
       }
    }
 }
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -124,7 +124,7 @@ namespace Bogus.Tests
       [Fact]
       public void generate_decimal_with_min_and_max()
       {
-         r.Decimal(2.2m, 5.2m).Should().Be(4.488022563614591414017291557m);
+         r.Decimal(2.2m, 5.2m).Should().Be(3.8697355489728032005903907232m);
       }
 
       [Fact]
@@ -176,12 +176,12 @@ namespace Bogus.Tests
       [Fact]
       public void generate_int32_many()
       {
-         r.Int().Should().Be(425714706);
-         r.Int().Should().Be(1767736486);
-         r.Int().Should().Be(-1019306602);
-         r.Int().Should().Be(-1398145763);
-         r.Int().Should().Be(1857743500);
-         r.Int().Should().Be(-1033498152);
+         r.Int().Should().Be(1077349347);
+         r.Int().Should().Be(1155054345);
+         r.Int().Should().Be(-1904480771);
+         r.Int().Should().Be(2101046113);
+         r.Int().Should().Be(1223601157);
+         r.Int().Should().Be(-594397672);
       }
 
       [Fact]
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -104,6 +104,96 @@ namespace Bogus.Tests
       }
 
       [Fact]
+      public void detects_invalid_Even_range()
+      {
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Even(min: 1, max: 0);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Even(min: int.MaxValue, max: int.MinValue);
+            });
+      }
+
+      [Fact]
+      public void detects_empty_Even_range()
+      {
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Even(min: 1, max: 1);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Even(min: int.MaxValue, max: int.MaxValue);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Even(min: int.MinValue + 1, max: int.MinValue + 1);
+            });
+      }
+
+      [Fact]
+      public void can_handle_extreme_Even_range()
+      {
+         r.Even(min: int.MinValue, max: int.MinValue).Should().Be(int.MinValue);
+         r.Even(min: int.MaxValue & ~1, max: int.MaxValue & ~1).Should().Be(int.MaxValue & ~1);
+      }
+
+      [Fact]
+      public void detects_invalid_Odd_range()
+      {
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Odd(min: 1, max: 0);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Odd(min: int.MaxValue, max: int.MinValue);
+            });
+      }
+
+      [Fact]
+      public void detects_empty_Odd_range()
+      {
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Odd(min: 0, max: 0);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Odd(min: int.MaxValue - 1, max: int.MaxValue - 1);
+            });
+
+         Assert.Throws<ArgumentException>(
+            () =>
+            {
+               r.Odd(min: int.MinValue, max: int.MinValue);
+            });
+      }
+
+      [Fact]
+      public void can_handle_extreme_Odd_range()
+      {
+         r.Odd(min: int.MinValue | 1, max: int.MinValue | 1).Should().Be(int.MinValue | 1);
+         r.Odd(min: int.MaxValue, max: int.MaxValue).Should().Be(int.MaxValue);
+      }
+
+      [Fact]
       public void random_bool()
       {
          r.Bool().Should().BeFalse();
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -91,10 +91,16 @@ namespace Bogus.Tests
       }
 
       [Fact]
-      public void exclusive_int_maxvalue_number()
+      public void can_include_int_maxvalue_number()
       {
-         var max = r.Number(int.MaxValue - 1, int.MaxValue);
-         max.Should().Be(int.MaxValue - 1);
+         var max = r.Number(int.MaxValue, int.MaxValue);
+         max.Should().Be(int.MaxValue);
+      }
+
+      [Fact]
+      public void can_handle_full_int_range()
+      {
+         r.Number(int.MinValue, int.MaxValue);
       }
 
       [Fact]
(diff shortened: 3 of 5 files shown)


Source/Bogus.Tests/GitHubIssues/Issue124.cs
```diff
@@ -129,24 +129,34 @@ public void nullable_int_and_nullable_reference_type()
             .RuleFor(x => x.Obj, f => new object().OrNull(f))
             .RuleFor(x => x.Str, f => f.Random.Word().OrNull(f));
 
-         var q = faker.Generate(3);
+         var q = faker.Generate(5);
```
logiclrd, line 132: Increased to 5 to ensure that at least one of the objects has a non-null `Str` value.

Source/Bogus.Tests/GitHubIssues/Issue260.cs
```diff
@@ -30,11 +33,12 @@ public void fast_algo3_test()
 
          // last 2^14 = 16384, for last 4 digits of SSN
          var c = (x >> 7) & 0x3FFF;
+         if( c >= 10000 ) c -= 10000;
```
logiclrd, line 36: Bug in test: `& 16383` does _not_ guarantee a 4 digit value.

Source/Bogus.Tests/RandomizerTest.cs
```diff
@@ -118,7 +124,7 @@ public void generate_double_with_min_and_max()
       [Fact]
       public void generate_decimal_with_min_and_max()
       {
-         r.Decimal(2.2m, 5.2m).Should().Be(4.0105668499183690m);
+         r.Decimal(2.2m, 5.2m).Should().Be(4.488022563614591414017291557m);
```
logiclrd, line 127: Increased range of decimal precision is evident.

bchavez:
Hi Jonathan,

First, thanks very much for the PR. I'm going to need some time to review everything but I'll get back to you on this.

Thanks,
Brian

Source/Bogus.Tests/GitHubIssues/Issue260.cs
bchavez: Actually, the whole algorithm for `fast_algo3_test()` for variables `a` and `b` are broken:
image
Good thing **ssn algo3** is not used in production.

IIRC, when I was testing issue 260, this was more of a test if I could squeeze more clock cycles out of the current SSN generator. It turned out, not much was to be gained going this far and ultimately stuck with the easier to understand implementation. :smile:

logiclrd: Ahh :-) So, for this PR, should I enhance `fast_algo3_test` to detect the errors and fix them, or just tear the test out entirely?

logiclrd: I mean, I guess the test isn't really testing a subject, all the logic is _in the test_, so if it's not being used ... tear it out?? It needs _some_ change because with the change to `Random.Number`, it no longer passes.

bchavez: Right on. For now, just making the test pass with the assertion should be enough. We can remove the bad algo separately from the PR; we just need this particular test to pass CI.

Currently benchmarking your new `Random.Number` with some alternatives. I'll post shortly.

logiclrd: Okay, then we're probably good then, right? Even if the algorithm is broken, the _test_ doesn't detect it presently, it only hits valid results. The test is green on my end.

bchavez: Yea, we're good. :+1:

bchavez:
## Random.Number() changes

Alright, so here are my initial thoughts change to `Random.Number()`. First, let's look at performance characteristics. Benched on `net471` and `netcoreapp3.1`:

```
BenchmarkDotNet=v0.12.1, OS=Windows 7 SP1 (6.1.7601.0)
Intel Core i7-3960X CPU 3.30GHz (Ivy Bridge), 1 CPU, 12 logical and 6 physical cores
Frequency=3224658 Hz, Resolution=310.1104 ns, Timer=TSC
.NET Core SDK=3.1.301
  [Host]        : .NET Core 2.1.19 (CoreCLR 4.6.28928.01, CoreFX 4.6.28928.04), X64 RyuJIT
  .NET 4.7.1    : .NET Framework 4.8 (4.8.3761.0), X64 RyuJIT
  .NET Core 3.1 : .NET Core 3.1.5 (CoreCLR 4.700.20.26901, CoreFX 4.700.20.27001), X64 RyuJIT
```

|    Method |           Job |       Runtime |     Mean |    Error |   StdDev |   Median | Allocated |
|---------- |-------------- |-------------- |---------:|---------:|---------:|---------:|----------:|
| OldMethod |    .NET 4.7.1 |    .NET 4.7.1 | 47.13 ns | 0.161 ns | 0.135 ns | 47.05 ns |         - |
|  BitShift |    .NET 4.7.1 |    .NET 4.7.1 | 56.40 ns | 0.126 ns | 0.105 ns | 56.36 ns |         - |
| JDGMethod |    .NET 4.7.1 |    .NET 4.7.1 | 58.91 ns | 0.316 ns | 0.280 ns | 58.76 ns |         - |
| OldMethod | .NET Core 3.1 | .NET Core 3.1 | 41.80 ns | 0.851 ns | 0.911 ns | 42.48 ns |         - |
|  BitShift | .NET Core 3.1 | .NET Core 3.1 | 52.16 ns | 0.033 ns | 0.031 ns | 52.17 ns |         - |
| JDGMethod | .NET Core 3.1 | .NET Core 3.1 | 53.11 ns | 0.049 ns | 0.039 ns | 53.10 ns |         - |

```csharp
[Benchmark]
public int OldMethod()
{
   return r.Number(int.MinValue, int.MaxValue);
}

[Benchmark]
public int BitShift()
{
   return r.NumberBitShift(int.MinValue, int.MaxValue);
}

[Benchmark]
public int JDGMethod()
{
   return r.NumberJDG(int.MinValue, int.MaxValue);
}
```

* `OldMethod` - Current production implementation.
* `JDGMethod` - Changes in this PR.
* `BitShift` - Slight modifications to this PR using a preallocated array to avoid calling `localSeed` twice:
```csharp
public int NumberBitShift(int min = 0, int max = 1)
{
   //lock any seed access, for thread safety.
   lock (Locker.Value)
   {
      // Adjust the range as needed to make max inclusive. The Random.Next function uses exclusive upper bounds.

      // If max can be extended by 1, just do that.
      if (max < int.MaxValue) return localSeed.Next(min, max + 1);

      // If max is exactly int.MaxValue, then check if min can be used to push the range out by one the other way.
      // If so, then we can simply add one to the result to put it back in the correct range.
      if (min > int.MinValue) return 1 + localSeed.Next(min - 1, max);

      localSeed.NextBytes(temp);
      return temp[0] << 24 | temp[1] << 16 | temp[2] << 8 | temp[3];
   }
}
```
### The Cost
Looking at the cost of `Random.Number()` changes:
* We are changing the contract of `Random.Number(_, int.MaxValue)`. Previously, we told users `int.MaxValue` was exclusive, and it will now be inclusive. Might break some people and their unit tests based on API contract change.
* The `Random.Number()` method would be +10ns more expensive. ~10ms per 1M generations.
* Deterministic sequences will change.

logiclrd:
So I just took a look into this, and it looks like the big performance hit is going to be _any_ call to `Random.Next(min, max)`, because what it does to produce the random number is significantly more expensive than what a caller probably does with that number.

Looking at the implementation, `Random.Next(min, max)` converts `Random.Next()` to a `double` and then back into the desired range. In most cases those conversions are liable to be more expensive than anything else the algorithm is doing, and I think that's definitely the case with Bogus.

This is the underlying source code:

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs

* `InternalSample` returns a value in the range `0` .. `int.MaxValue - 1`. At first, I thought the last value is 1.5 times as likely as any of the others because of the bizarre decision to convert `int.MaxValue` into `int.MaxValue - 1`, but after testing it, the last value is less likely than `0` or `0x12345678`, so I guess the underlying generator isn't perfectly even. :-P (It also looked like it might occasionally return `-1`, because of the line `if (retVal < 0) retVal += MBIG;`, but it turns out `retVal` can't ever be exactly `int.MinValue` at that line because of how the generator works.)

* `Next()` just straight up returns this number:
```
        public virtual int Next()
        {
            return InternalSample();
        }
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs#L160-L163

* But `Next(min, max)` converts it to a `double`, scales that, then converts it back to an `int`:
```
                return (int)(Sample() * range) + minValue;
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs#L201
```
        protected virtual double Sample()
        {
            // Including this division at the end gives us significantly improved
            // random number distribution.
            return InternalSample() * (1.0 / MBIG);
        }
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs#L87-L92

* And that's for small ranges. If the range has more than `int.MaxValue` possible results, then it actually pulls two random values and combines them:
```
            if (range <= int.MaxValue)
                (..)
            else
            {
                return (int)((long)(GetSampleForLargeRange() * range) + minValue);
            }
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs#L205
```
            int result = InternalSample();
            // Note we can't use addition here. The distribution will be bad if we do that.
            bool negative = (InternalSample() % 2 == 0) ? true : false;  // decide the sign based on second sample
            if (negative)
            {
                result = -result;
            }
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Random.cs#L172-L178

This combined value is then turned into a `double`.

Based on this, the only way I can see to significantly improve performance is to call `Next()` directly instead of `Next(int.MinValue, 0)` and combine individual bits from the results, avoiding the floating point conversions. `Next()` gives slightly less than 31 bits of random data on each call, so two calls to it is obviously more than enough to get 32 bits of random data.

Based on this, I propose the following algorithm:

```
    int sample1 = localSeed.Next();
    int sample2 = localSeed.Next();

    int topHalf = (sample1 >> 8) & 0xFFFF;
    int bottomHalf = (sample2 >> 8) & 0xFFFF;

    return unchecked((topHalf << 16) | bottomHalf);
```

(I ignore the lowest and highest bits in each case, as I suspect they may have less entropy than the middle bits.)

I ran a test of this to verify distribution, left it running on a bunch of nodes for a while, and after 2.37 trillion samples, 552 had been `0`, 531 had been `int.MaxValue - 1`, 557 had been `int.MaxValue` and 542 had been `0x12345678`. The proportion that were negative came out to 0.500000204.

How does this algorithm perform side-by-side with the other implementations?

logiclrd:
I decided to pull 10,000 samples and graph them, on the supposition that if there were patterns in it, they'd probably stand out to my eye.

image

bchavez:
Hi Jonathan, thanks for looking further into the situation. I don't think the +10ns is a problem per se, I just needed a general understanding of the perf characteristics and other costs associated with the PR. This is a core change to `Randomizer` so I'm being extra careful.

I do really like your idea of subtracting -1 and adding +1 to the result. It's pretty clever. Kinda wish we had this earlier during the beginning.

The biggest concern for me right now is removing the upper bound limit of `int.MaxValue - 1` and the implementation change in `Number`; but it's slowly growing on me. As I see it now, `Randomizer.Int()` is currently broken because `Randomizer.Number()` never returns `int.MaxValue`.

As for your newest method `JDGMethod2`:

https://github.com/bchavez/Bogus/blob/59c17aa248735864b5a26e6a4ee4d8ffa4e6b9b0/Source/Benchmark/BenchRandomNumber.cs#L114-L129

Benchmark BenchRandomNumber-barplot

|     Method |       Runtime |     Mean |    Error |   StdDev | Allocated |
|----------- |-------------- |---------:|---------:|---------:|----------:|
|  OldMethod |    .NET 4.7.1 | 46.75 ns | 0.046 ns | 0.041 ns |         - |
|   BitShift |    .NET 4.7.1 | 58.00 ns | 0.071 ns | 0.059 ns |         - |
|  JDGMethod |    .NET 4.7.1 | 57.66 ns | 0.170 ns | 0.142 ns |         - |
| JDGMethod2 |    .NET 4.7.1 | 37.68 ns | 0.022 ns | 0.019 ns |         - |
|  OldMethod | .NET Core 3.1 | 39.34 ns | 0.034 ns | 0.028 ns |         - |
|   BitShift | .NET Core 3.1 | 52.06 ns | 0.067 ns | 0.060 ns |         - |
|  JDGMethod | .NET Core 3.1 | 53.27 ns | 0.292 ns | 0.228 ns |         - |
| JDGMethod2 | .NET Core 3.1 | 33.67 ns | 0.047 ns | 0.036 ns |         - |

I hope to finish my review tomorrow. Again, thanks so much for being responsive as I go through this review.

logiclrd:
Nice! Thanks for plugging it in :-)

Source/Bogus/Randomizer.cs
```diff
@@ -75,46 +75,79 @@ public int[] Digits(int count, int minDigit = 0, int maxDigit = 9)
...
+         // Adjust the range to ensure that we always get the same number of even values as odd values.
+         min = (min + 1) & ~1;
+         max = max | 1;
+
+         if( min > max )
+            return min;
+
+         // Strip off the 1 bit of a random number.
+         return Number(min, max) & ~1;
       }
```
bchavez, line 129: I'm thinking we should probably throw an exception if the user gets it wrong.

For example, if the user calls `.Even(min:3, max:3)` there are no even numbers; however, in this change, we would return 4 which seems unsuspecting. The user might not be aware of their mistake.

Kinda feel, if I call `.Even(min, max)` I should get exactly that, between `min` and `max`. Semantics wise, I would not expect a value outside of `min` and `max`; even if I made the mistake on purpose.

I assumed when I wrote many of these `Randomizer` methods that the inputs would be correct and if the user got it wrong, "oh well". I suppose if we're going to get proper about it, I think we can lock down these out of bounds arguments with exception guards.

Let me know your thoughts.

bchavez: LoL, the production version in `master` enters an infinite loop when there are no `.Even` numbers:
```csharp
void Main()
{
   var f = new Faker();
   f.Random.Even(3, 3).Dump();
}
```
So I guess, the current behavior is that we've already broken the user's program and they might have noticed this when Bogus got stuck.

logiclrd:
There is some subtlety to the range check in `Even` and `Odd`. Not sure if it warrants explicit documentation. Basically, after determining that `min` is an odd number, `Even` then checks if the range has an even number by testing if `max - 1` is still in range. It doesn't check `min + 1`, because if `min` is an odd number, it _might_ be equal to `int.MaxValue`. Same logic in reverse for `Odd`.

The unit tests I added do exercise the extreme values explicitly.

bchavez:
Hi Jonathan, thanks so much for adding those unit tests and for adding the argument guards. I think we're pretty close to crossing the finishing line now.

For the last part of this PR, I benched the new `.Decimal()` method and it looks like we got a bump in allocations in **Gen 0** probably due to the [`int[] bits = new int[4];`](https://github.com/bchavez/Bogus/pull/300/files#diff-967f5ae0cb5988159caeb1388486d113R210). Here are the perf characteristics for the new `Random.Decimal()`:

* `OldMethod` - in `master`
* `JDGMethod` - in this `.Decimal()` in this PR.
* `JDGMethodNoAlloc` - Slightly modified version with no allocation using `new decimal(lo, mid, hi, sign, scale)` constructor.
```csharp
public decimal DecimalJDGNoAlloc(decimal min = 0.0m, decimal max = 1.0m)
{
   int lo = NumberJDG(int.MinValue, int.MaxValue);
   int mid = NumberJDG(int.MinValue, int.MaxValue);
   int hi = NumberJDG(int.MinValue, int.MaxValue);
   byte scale = 0x1C;

   decimal result = new decimal(lo, mid, hi, false, scale);

   return result * (max - min) / 7.9228162514264337593543950335m + min;
}
```

Benchmark BenchDecimal-barplot

|           Method |       Runtime |      Mean |    Error |   StdDev |  Gen 0 | Allocated |
|----------------- |-------------- |----------:|---------:|---------:|-------:|----------:|
|        OldMethod |    .NET 4.7.1 | 221.39 ns | 0.320 ns | 0.250 ns |      - |         - |
|        JDGMethod |    .NET 4.7.1 | 522.91 ns | 1.796 ns | 1.593 ns | 0.0057 |      40 B |
| JDGMethodNoAlloc |    .NET 4.7.1 | 519.23 ns | 0.685 ns | 0.572 ns |      - |         - |
|        OldMethod | .NET Core 3.1 |  84.27 ns | 0.669 ns | 0.593 ns |      - |         - |
|        JDGMethod | .NET Core 3.1 | 321.60 ns | 1.779 ns | 1.577 ns | 0.0048 |      40 B |
| JDGMethodNoAlloc | .NET Core 3.1 | 308.96 ns | 1.105 ns | 0.923 ns |      - |         - |

The bump in runtime is probably okay - we're getting more accuracy, but perhaps we can get back down to zero allocation with the `new decimal(lo, hi, mid,...)` constructor? Let me know what you think.

There's a power outage in my area and I'm currently running on UPS battery backup, so responses will be a little slow. I hope to get back online soon once utility power is back online.

logiclrd:
This seems sensible :-)

One possible optimization, if we don't mind the upper bound being _technically_ exclusive, would be to multiply by the reciprocal. The reciprocal isn't exactly expressible in a `decimal` -- in my tests, the average of the closest number below the reciprocal and the closest number above the reciprocal is essentially equal to the reciprocal at `decimal`'s level of accuracy. But, if we don't mind a call to `Decimal(0m, 1m)` having a maximum return value of `0.9999999999999999999999999996m`, then instead of dividing by `7.9228162514264337593543950335m`, multiply by `0.1262177448353618888658765705m`. In my testing, even including the 6 calls to `Random.Next()`, the execution time decreased by nearly 40%.

I don't think anybody will actually be writing tests that require `Decimal(0m, 1m)` to ever actually return exactly 1m, as that would only be expected to happen in one out of every 2^96 test runs anyway. :-P

bchavez:
Hi Jonathan, I'm okay with either approach as long as we do our best to knock out the **Gen 0** allocation.

I'm okay with an exclusive `1m` upper bound. We got to 5.7M downloads without major problems even with all issues currently in `master` that we're just now fixing. :) If we can get a decrease in execution time with a little bit of imperfection, it's probably worth a shot.

Ultimately, though, you can decide.

bchavez:
If you do change the PR, I'll do one last perf check on the latest changes.

logiclrd:
Hmm, I'm of two minds on this. :-P On the one hand, it is a substantial performance difference, but on the other, the cost of technical correctness is that we can only generate 2 million bogus `decimal` values per second? That seems like a lot. I don't know what all consumers Bogus has. I can guess that there may be edge cases where immense numbers of values are needed, but for sure the _average_ case, this performance difference is lost in the noise.

I'm inclined to keep it correct, do the division, personally. Let me know if you feel differently :-)

bchavez:
Sounds good. I'll add a few more code comments on top of this PR for documentation that I found helpful during this review. I should be able to carry this across the finish line now.

Also, I think we'll release this under a `v30.0.1-beta-1` tag, and let it be consumed in the wild for a little bit. We'll promote it to a full `v30` release after it's had some time to be digested by the community just in case we missed something in the review.

Thanks again Jonathan for this PR. You did a great job here. When you get a chance, send me an email bchavez@bitarmory.com and I'll send you a small gift as appreciation for helping Bogus and the community with this PR. Be sure to post the hash of your email so I know it's you. Alternatively, you can also contact me on [keybase](https://keybase.io/bchavez) which shows your verified GitHub account.

logiclrd:
Well thank you! This has been a very pleasant experience from my end, too :-)

I must admit to not being familiar with either of those things. :-) I've never used keybase, and while I'm perfectly comfortable hashing arbitrary byte sequences, I'm not quite sure how to do that to an e-mail in a way that ensures that you're hashing the same thing I am. There's probably some standard for it that I don't know about. :-)

bchavez:
No problem. Send an email. I'll send you a GUID, post that GUID here in GH, and I'll know it's you.

logiclrd:
I received the following GUID by e-mail: `2344f58c-c8c1-4d3a-8b7c-d463eb6d3413`

:-)

PR #303, opened 2020-06-20
Make Between and BetweenOffset in Date class optional

weitzhandler:
This PR makes the parameters of `Between` and `BetweenOffset` optional, so we can choose to omit them and use system min or max value.

### DIFF
--- a/Source/Bogus.Tests/DataSetTests/DateTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/DateTest.cs
@@ -181,6 +181,33 @@ namespace Bogus.Tests.DataSetTests
       }
 
       [Fact]
+      public void can_get_random_time_between_two_dates_optional()
+      {
+         var now = DateTime.Now;
+
+         //not specified
+         date.Between()
+            .Should()
+            .BeOnOrAfter(DateTime.MinValue)
+            .And
+            .BeOnOrBefore(DateTime.MaxValue);
+
+         //only specify start
+         date.Between(start: DateTime.Now)
+            .Should()
+            .BeOnOrAfter(now)
+            .And
+            .BeOnOrBefore(DateTime.MaxValue);
+
+         //only specify end
+         date.Between(end: DateTime.Now)
+            .Should()
+            .BeOnOrAfter(DateTime.MinValue)
+            .And
+            .BeOnOrBefore(now);
+      }
+
+      [Fact]
       public void can_get_random_time_between_two_dateOffsets()
       {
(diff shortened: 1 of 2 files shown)


bchavez:
Hi @weitzhandler, thank you for the pull-request. I appreciate it very much; however, I don't agree with the public API change here. The reason is the call semantics are somewhat confusing. Let us consider the following code:

```csharp
void Main()
{
   var startDate = DateTime.Parse("11/1/2020");
   var endDate = DateTime.Parse("12/1/2020");
   
   var userFaker = new Faker<User>()
      .RuleFor(u => u.CreatedDate, f => f.Date.Between(startDate, endDate)) //clear
      .RuleFor(u => u.ModifiedDate, f => f.Date.Between(end:endDate)); //hm...
}

public class User
{
   public DateTime CreatedDate{get;set;}
   public DateTime ModifiedDate{get;set;}
}
```
At first glance, when I read `f => f.Date.Between(end:endDate)` it does not make sense to me. I get the feeling I need to look at the implementation of `Between(end:endDate)` to make sure I know what I'm getting if I were to call this method with this particular argument. Maybe XML documentation can help, but for me, that's a bad first step.

-----

If you need support for this API scenario or prefer the ergonomics of default values, consider creating an extension method for the `Bogus.DataSet.Date` data set with your particular API surface preferences:

```csharp
void Main()
{
   var startDate = DateTime.Parse("11/1/2020");
   var endDate = DateTime.Parse("12/1/2020");

   var userFaker = new Faker<User>()
      .RuleFor(u => u.CreatedDate, f => f.Date.Between(startDate, endDate))
      .RuleFor(u => u.ModifiedDate, f => f.Date.GetBefore(endDate));
}

public class User
{
   public DateTime CreatedDate { get; set; }
   public DateTime ModifiedDate { get; set; }
}

public static class ExtensionsForDate
{
   public static DateTime GetBefore(this Bogus.DataSets.Date dt, DateTime beforeDate)
   {
      return dt.Between(DateTime.MinValue, end: beforeDate);
   }
}
```
-----
Also, if you'd like to change any public APIs please make an issue for the suggested enhancement. This would give us a chance to discuss your proposed changes. Also, this way there'd be a greater chance of getting a PR merged.

I appreciate the effort very much but I'm not totally sold on the public API change in this PR.

weitzhandler:
Now that's a quick response. I really appreciate the time and effort you took to put this detailed response together!
How about a new method `Any` or `AnyOffset` that has no start and end constraints at all?

weitzhandler:
Thank you so much for your quick and detailed response. I understand your concern.
How about a new method with a different name, like `PastOrFuture`, `Unrestricted` or anything of that kind?

PR #322, opened 2020-09-04
Improving README.md docs: Without Fluent Syntax

ZackStone:
I split the example code into 3 blocks, so it's easier to read.


bchavez:
Thank you for the help @ZackStone!

PR #338, opened 2020-11-23
UseConstructor(), SkipConstructor() and fixing StrictMode() error

Basyras:
➕  Adding features for specifying constructors: `UseConstructor()` and `SkipConstructor()`
🔧  Found & Fixed bug when StrictMode(true) is used and no RuleFor() is specified

⚠️  .net standard 1.3 is missing few methods so `NotSupportedException` is thrown istead in this assembly

### DIFF
--- a/Source/Bogus.Tests/FakerTests.cs
+++ b/Source/Bogus.Tests/FakerTests.cs
@@ -1,0 +1,237 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Bogus.DataSets;
+using Bogus.Extensions;
+using Bogus.Tests.Models;
+using FluentAssertions;
+using Xunit;
+
+
+namespace Bogus.Tests
+{
+   public class FakerTests
+   {
+      [Fact]
+      public void SkipConstructor_Default_StrictMode_Satisfied()
+      {
+         _ = new Faker<NoParameterlessCtorClass>()
+            .SkipConstructor()
+            .RuleFor(x => x.Obj, x => null)
+            .RuleFor(x => x.Obj2, x => null)
+            .RuleFor(x => x.Obj3, x => null)
+            .Generate(1);
+      }
+
+      [Fact]
+      public void SkipConstructor_WithRule_Default_StrictMode_Not_Satisfied()
+      {
+         Assert.Throws<ValidationException>(() =>
+         {
+            _ = new Faker<NoParameterlessCtorClass>()
+               .SkipConstructor()
+               .RuleFor(x => x.Obj, x => null)
+               .Generate(1);
+         });
+      }
+
+      [Fact]
+      public void SkipConstructor_WithoutRules_Default_StrictMode_Not_Satisfied()
+      {
+         Assert.Throws<ValidationException>(() =>
+         {
+            _ = new Faker<NoParameterlessCtorClass>()
+               .SkipConstructor()
+               .Generate(1);
+         });
+      }
+
+      [Fact]
+      public void Not_Skipping_Constructor_Without_Parameterless_Ctor()
+      {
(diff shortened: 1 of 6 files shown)


Source/Bogus.Tests/FakerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            _ = new Faker<NoParameterlessCtorClass>()
+                  .StrictMode(true)
+                  .UseConstructor(x => new NoParameterlessCtorClass(default, default))
+                  .Generate(1);
+         });
+      }
+
+
+      [Fact]
+      public void UseContructor_After_SkipConstructor()
```
Basyras, line 128: Is debatable if using both `UseConstructor` and `SkipConstructor` should be prohibited or not although it does not break anything.

Source/Bogus/Faker[T].cs
```diff
@@ -159,16 +176,108 @@ public virtual Faker<T> UseSeed(int seed)
...
+      /// </summary>      
+      /// <returns></returns>
+      public virtual Faker<T> SkipConstructor()
+      {
+#if NETSTANDARD1_3
+         throw new NotSupportedException($"{nameof(SkipConstructor)}() is not supported in .net standart 1.3. Consider using {nameof(UseConstructor)}().");
+#else
+         //Not sure wich works better
+         //this.CreateActions[Default] = (Faker faker) => (T)FormatterServices.GetUninitializedObject(typeof(T));
+         UseConstructor((Faker faker) => (T)System.Runtime.Serialization.FormatterServices.GetUninitializedObject(typeof(T)));
```
Comment (Basyras) [inline Source/Bogus/Faker[T].cs:268]: Not sure if this.CreateActions[Default] should be called or this.CreateActions[currentRuleSet] instead.

Source/Bogus/Faker[T].cs
```diff
@@ -740,14 +863,14 @@ private ValidationResult ValidateInternal(string[] ruleSets)
...
             //that the user does not know about + .Rule() methods.
 
-            if( userSet.Count > 0 )
+            if (userSet.Count > 0)
             {
-               foreach( var propOrFieldOfT in userSet )
+               foreach (var propOrFieldOfT in userSet)
                {
-                  if( populateActions.TryGetValue(propOrFieldOfT, out var populateAction) )
+                  if (populateActions != null && populateActions.TryGetValue(propOrFieldOfT, out var populateAction))
```
Comment (Basyras) [inline Source/Bogus/Faker[T].cs:870]: Mentioned bug was here when no RuleFor() was called (Or simple AddRule) when StrictMode(true) the populateActions  variable was null and resolved in null refernce exception

bchavez:
@Basyras Thank you for the PR. I really appreciate it! I'll try to provide my thoughts below.

* Please try to revert formatting changes. Please ***only*** change what is **minimally** necessary. Reviewing this PR is extremely difficult because of the formatting changes. Code merged into Bogus should be production quality and goes through a strict review process because, ultimately, I have to maintain it. 

* Keep the code focused on the new feature only. That is, the PR should only be adding `.SkipConstructor()` that uses `GetUninitializedObject`.

* If you see a bug, the bug fix should be in a separate PR, not merged with a new feature PR. It's too difficult to track and revert if there is a problem with the bug and fix.

If you can, please break up the PR. There are too many changes in this PR. I will try to dedicate more time for this during the holiday break.

Thanks,
Brian

Basyras:
@bchavez Shared formatting rules from .editorconfig file were not applied. I can see .editorconfig file in your root folder but i think your formatting rules are not inside. That's the reason why my formatting changes were made. Could you fix that? Should be simple, go to your IDE formatting settings and on top there should be option to generate .editorconfig

gl1tch:
What ever happened to this PR, I don't seem to see it ever made it into the code ?

Basyras:
@gl1tch nope

PR #350, opened 2021-01-21
The Summary Is Misleading

M-Clk:
The summary is incorrect. # defines the numeric characters and ? defines a letter. In summary it is the opposite :)


bchavez:
Thank you @M-Clk! :+1: :sweat_smile:

PR #352, opened 2021-01-31
Fix first name generation for locales that don't support gender names

MZabal:
When instantiating a new Person for the spanish locale the generated name was always english. 
I have added condition to return First Names from the default list if the locale doesn´t support gendered names


bchavez:
Great work! Thank you very much for adding a unit test with your PR. :100: New release pending shortly.

bchavez:
@MZabal new version released with your PR:
https://www.nuget.org/packages/Bogus/32.1.1

Thanks again for the unit test!

PR #359, opened 2021-02-18
add swedish personumber

klowdo:
This adds generation of the swedish personumber

### DIFF
--- a/Source/Bogus/Extensions/Sweden/ExtensionsForSweden.cs
+++ b/Source/Bogus/Extensions/Sweden/ExtensionsForSweden.cs
@@ -1,0 +1,89 @@
+using System;
+using System.Linq;
+using Bogus.DataSets;
+
+namespace Bogus.Extensions.Sweden
+{
+   /// <summary>
+   /// API extensions specific for a geographical location.
+   /// </summary>
+   public static class ExtensionsForSweden
+   {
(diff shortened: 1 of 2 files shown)


Source/Bogus/Extensions/Sweden/ExtensionsForSweden.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+   /// <summary>
+   /// API extensions specific for a geographical location.
+   /// </summary>
+   public static class ExtensionsForSweden
+   {
+      /// <summary>
+      /// Swedish national identity number (personummer)
+      /// </summary>
+      public static string Personummer(this Person person)
```
bchavez, line 15: Should the name of this extension method be `Personnummer` with two `nn`s?
https://en.wikipedia.org/wiki/Personal_identity_number_(Sweden)

klowdo: You are absolutely right! thanks i'll fix

klowdo: fixed

bchavez:
@klowdo, changes released in https://www.nuget.org/packages/Bogus/33.0.2
Thanks for your contribution!

Issue #365, opened 2021-03-16
Bogus can generate impossible DateTime values


logiclrd:
### Version Information
| Software                       | Version(s) |
| ------------------------| ---------- |
| Bogus NuGet Package  | 31.0.3, 33.0.2    |
| .NET Core?                   |                 |
| .NET Full Framework?   | 4.8             |
| Windows OS?              | 10               |
| Linux OS?                    |                  |
| Visual Studio?              | 16.9.0 Preview 5 |

### What locale are you using with Bogus?
en-US

### What is the expected behavior?
Generated `DateTime` values _always_ represent realistic dummy data.

### What is the actual behavior?
If the range of accepted values contains a Daylight Savings Time shift forward, sometimes Bogus can return impossible `DateTime` values. For instance, in my locale, dates between 2021-03-14 02:00:00 and 2021-03-14 02:59:59.9999999 don't normally exist, and supplying such values to formatting and serialization functions can cause unexpected behaviour. Bogus has caused automated tests to fail by returning such values for dummy data.

### Please provide a stack trace.
n/a

### Any possible solutions?
Detect and correct invalid values at every site that asks Bogus for a DateTime value. But, sometimes these sights are buried inside other libraries (e.g. AutoBogus).

### How do you reproduce the ?
Ensure that the range requested includes impossible DateTime values and request enough values and eventually you'll get an impossible value.

One way in which these invalid values can cause real-world problems is in serialization. For instance, if an invalid value is put into a `DataTable` object, then when that `DataTable` is serialized and then deserialized, the deserialized copy will have a different `DateTime` value. In my usage so far, this is mostly a problem in automated testing, and automated testing is where Bogus is used as well.

### Do you have a unit test that can demonstrate the bug?
```
			var faker = new Faker();

			const int SampleCount = 1_000_000;

			int badSampleCount = 0;

			var firstInvalidDateTime = new DateTime(2021, 3, 14, 2, 0, 0);
			var firstValidDateTime = new DateTime(2021, 3, 14, 3, 0, 0);

			for (int i = 0; i < SampleCount; i++)
			{
				var value = faker.Date.Between(new DateTime(2021, 3, 14), new DateTime(2021, 3, 15));

				if ((value >= firstInvalidDateTime) && (value < firstValidDateTime))
					badSampleCount++;
			}

[...]

PR #366, opened 2021-03-16
Update the Date data set to avoid DST transition windows

logiclrd:
This PR:

* Adds a unit test that I'm _pretty sure_ will find a DST window and force a value to generate inside of it. Works on my system, anyway. The test should automatically skip if the local time zone doesn't do DST transitions.
* Updates the implementation of the `Date` data set so that all requests get funnelled into `Between` and `BetweenOffset`, and then reimplements `Between` and `BetweenOffset` to convert the range to UTC first, then obtain a random value, and then convert it back to a local time if appropriate before returning it, taking advantage of the framework's daylight savings transition calculations.

The unit test fails with the original `Date` data set code, and the change in the `Date` data set fixes the unit test. On my system, at least.

This PR is in response to #365.


logiclrd:
Rebased. I thought I was working with the latest `master`, but I wasn't. The change is now in terms of the latest `master`.

Source/Bogus/DataSets/Date.cs
```diff
@@ -161,14 +153,19 @@ public DateTime Between(DateTime start, DateTime end)
...
+         var maxTicks = Math.Max(startTicks, endTicks);
 
          var totalTimeSpanTicks = maxTicks - minTicks;
 
          var partTimeSpan = RandomTimeSpanFromTicks(totalTimeSpanTicks);
 
-         return new DateTimeOffset(minTicks, start.Offset) + partTimeSpan;
+         var dateTime = new DateTime(minTicks, DateTimeKind.Unspecified) + partTimeSpan;
+
+         return new DateTimeOffset(dateTime + start.Offset, start.Offset);
```
bchavez, line 168: I'm going to replace these two lines with:

```csharp
         if( start < end )
            return start + partTimeSpan;
         else
            return end + partTimeSpan;
```

Given an invalid `start` time; the function will still return an invalid `DateTimeOffset`.

image

With the code change, at least the method returns a valid time, albeit outside the "specified" time range; which is fine I guess, because as it stands now, it would still have been outside the "2:59 AM - 3:00 AM" time as noted above.

image

logiclrd: That seems legit :-)

logiclrd: I guess to my mind, if the caller supplies impossible DateTime values to begin with, all bets are off. My specific use case doesn't intersect with that, so I don't have strong opinions about how exactly it behaves in that edge case.

logiclrd: Thinking it through further, if one wanted to be _really_ thorough, then the algorithm would need to include some sort of "compute _real_ DateTime range" functionality, so that `start` and `end` were adjusted if they lay inside an impossible time span, and this would then also throw if `start` and `end` both lay within the same transition window (so that no real date/time `x` would satisfy `start <= x <= end`). This would be in addition to generating the actual bogus value in UTC, to avoid transition windows in the middle of the range. I think this would be the most polished form of this, but I'm not sure if it's worth the effort.

logiclrd: Pseudocode:
```
void ComputeRealRange(ref DateTime start, ref DateTime end)
{
  if (start > end)
    Swap(ref start, ref end);

  var (windowStart, windowEnd) = GetDSTTransitionWindow(start.Year);

  if (windowStart <= start <= windowEnd)
    start = windowEnd;

  (windowStart, windowEnd) = GetDSTTransitionWindow(end.Year);

  if (windowStart <= end <= windowEnd)
    end = windowStart;

  if (start > end)
    throw new Exception("Impossible");
}
```

logiclrd: With this approach, a supplied range as shown where `start` lay within the impossible range and `end` was exactly the first next valid `DateTime` value, it would always return exactly the `end` value because the _effective_ `start` would be bumped to the end of the transition window and would coincide with `end`.

logiclrd:
Any changes you'd like to request on this PR? :-)

logiclrd:
I have implemented the proposed algorithm for excluding impossible values from the range by bumping `start` forward and `end` back, along with unit tests that detect the correction.

logiclrd:
There were some shenanigans surrounding `Kind` and offsets but that's now sorted out. For your test, with `start` set to 1 minute before the end of the transition and `end` set to exactly the end of the transition, the requested range actually only has one real `DateTime` value, that being the exact end of the transition, and so it will always return that exact value.
```
2021-03-14 2:59:00 AM -06:00
2021-03-14 3:00:00 AM -05:00
2021-03-14 8:59:00 AM +00:00
2021-03-14 8:00:00 AM +00:00
2021-03-14 3:00:00 AM -06:00
```

logiclrd:
Sorry, I don't normally do so many force-pushes, but in my eagerness I'm afraid I pushed changes only to realize a few minutes later that I had overlooked a case and return to the code to address it.

bchavez:
@logiclrd no worries. Thank you for making the improvements; sorry for the delay getting this merged. I've been busy with a few life events that are taking a bit of time to resolve.

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +353,236 @@ public void can_get_timezone_string()
       {
          date.TimeZoneString().Should().Be("Asia/Yerevan");
       }
+
+      public class FactWhenDaylightSavingsSupported : FactAttribute
+      {
+         public FactWhenDaylightSavingsSupported()
+         {
+            if (!TimeZoneInfo.Local.SupportsDaylightSavingTime)
```
Chobocharles, line 361: Wow, I had no clue this was a thing.

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +353,236 @@ public void can_get_timezone_string()
...
+
+         faker.Random = new Randomizer(localSeed: 5);
+
+         var dstRules = TimeZoneInfo.Local.GetAdjustmentRules();
+
+         var now = DateTime.Now;
+
+         var effectiveRule = dstRules.Single(rule => (rule.DateStart <= now) && (rule.DateEnd >= now));
+
+         var transitionStartTime = CalculateTransitionDateTime(now, effectiveRule.DaylightTransitionStart);
```
Chobocharles, line 382: New to this code base, but what's the consensus on generating test helper methods or classes that can contain boiler plate construction and arrangement of objects?

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +353,236 @@ public void can_get_timezone_string()
...
```
Chobocharles, line 411: Because we can probably toss all of this initial setup into a base method (except for the rule) and just call the method at the beginning of the tests

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: My convention when writing tests has been to make the test methods completely logically independent of one another, but I have seen tests and test suites that reuse a great deal of setup. This could be refactored, but was written this way intentionally (if without a great deal of thought, per se).

logiclrd:
@bchavez Ping :-)

bchavez:
Hey @logiclrd , im still here :) I'll try to give it a shot this weekend!!

logiclrd:
@bchavez How's it going? :-D

bchavez:
hey @logiclrd , for sure, I need to get back to this - I just haven't found enough time to do the extensive testing I want to do before I merge this PR

bchavez:
@logiclrd , it was the same test I did that found the previous issue in the PR here:
* https://github.com/bchavez/Bogus/pull/366#discussion_r598342181

I know it is probably fixed, but I need to see it first-hand and probably see how it behaves in other scenarios too.

logiclrd:
The code has changed since then. I'll run the same test, see what happens. :-)

logiclrd:
Okay, I added a test that I think encapsulates what you were talking about, and I tweaked the behaviour of the `Date` set to match the expectations encoded in the tests.

logiclrd:
Let me know if that test covers the functionality you were thinking of. :-)

bchavez:
Thank you @logiclrd -- I appreciate it; hopefully, after I get personal and business taxes out of the way I can take a 2nd look. Just really difficult to do OSS work after a full-time job; and I get pretty tired by the end of the day.

There's always a risk; because as soon as I merge and release this code, I have to be ready for an influx of support requests which is why I'm waiting until I have enough buffer-time to handle any immediate issue that could come up as a result of some **_new_** time code bug that could come up as a result of this new code.

Yes, we do have a bug now, but at least it's known -- and hasn't caused any major problems as far as GH issues are concerned.

logiclrd:
That's fair :-) For what it's worth, we've been running on a private prerelease build on our private Azure Artifacts server for, well, for some time now :-P

logiclrd:
How's it going? :-)

logiclrd:
Where are we at?

voronoipotato:
Just hit this trying to test an issue I was having where the unit test was working locally but failing on the build server configured to use UTC. 
```F#
      Expected: ···2022-03-29 05:46:04Z
      Actual:   ···2022-03-29 05:46:03Z 
```

building this test was actually quite difficult, because the faker facade really really doesn't want to let go of the seed even though I created a whole new instance. Turns out you have to set the seed on the original faker instance and not the new one, but you also do need to create a new instance, and you need to do it before you update the seed. 

```F#
System.Threading.Thread.CurrentThread.CurrentCulture <- CultureInfo.InvariantCulture
let faker =Faker(Random = (Randomizer(777), Localization = "en")
let data = generateDates faker //some code to make a ton of dates
let faker2 =Faker(Random = (Randomizer(777), Localization = "en")
faker.Random <- Randomizer(777) 
//if I don't do this for some reason faker produces entirely different datasets even though faker2 should be an entirely different instance, also setting faker.Random on faker 2 does nothing at all. 
let data2 = generateDates faker2
Assert.IsTrue(data.ToString() = data2.ToString())
```

probably going to create a new ticket for this since it can make it nearly impossible to replicate bugs without determinism, and it would be nice if separate faker instances were at least "able" to be completely separate.

logiclrd:
@bchavez Any chance of getting this loose end tied off? :-)

logiclrd:
Oh, you know what, I was mistaken. We don't have a custom build of Bogus. We have a custom build of FluentAssertions with support for System.Data types. We just still get random test failures around DST transition. Another developer at some point made it retry tests up to 5 times to handle stochastic failures. That's its own thing :-P

logiclrd:
Maybe we can get this in place before the next DST switch? :-)

logiclrd:
The next DST switch is in 5 days.

logiclrd:
Well, I guess the next DST switch _that matters_ is March 12, 2023. :-P

logiclrd:
@bchavez Ping :-)

logiclrd:
10 days away!

logiclrd:
Hello from the other side :-)

bchavez:
Hey @ChristoWolf, if you have a chance or any free time;

Could you double-check this PR and see if we have sufficient unit test coverage for edge cases? I'd like to know your thoughts on this PR. 

I'm curious if there are any edge cases where this new code might break existing code. Or ways we could intentionally break the new code or its assumptions.

Right now, currently working on fixing the CI/CD build system and getting the pipeline up and running again. After that I'll likely switch gears to this PR; but would be nice to have a fresh set of eyes and opinions on it.

logiclrd:
I'm not saying that it's impossible that there might yet be bugs, but I don't _think_ they are, and the _very purpose_ of the reworked code is to completely rule out edge cases with impossible dates by design. If you can find a way to break assumptions with the new code, though, that'll be good :-) Better to find it out now!

ChristoWolf:
@bchavez: Sure, I'd love to, hope I can do it already today!

ChristoWolf:
@bchavez: Not sure why, but I am having big troubles with building the solution, fake is messing up no matter which version I use.
I don't think I had this issue in the past.
In any case, I will for now review the PR without running the changes.

bchavez:
@ChristoWolf ah, for sure, the F# FAKE stuff is busted right now; I think the root cause is mostly in `\Source\Builder\build.fsx.lock` files that attempt to get resolved, but are fundamentally not compatible with the latest .NET 7 SDK.

As a workaround, I think you might be able to rely on the base CLI commands like `dotnet restore; dotnet build; dotnet test` in the `\Source` folder. Also, I think opening the SLN and building/testing should work from within the IDE itself too.

Let me know if there's anything else I can try to help with.

ChristoWolf:
@bchavez: Good to know that it's not just me :)
Got it working, thanks!

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
       /// <param name="end">End time</param>
       public DateTime Between(DateTime start, DateTime end)
       {
-         var minTicks = Math.Min(start.Ticks, end.Ticks);
-         var maxTicks = Math.Max(start.Ticks, end.Ticks);
+         ComputeRealRange(ref start, ref end, start.Kind, out var preferRangeBoundary);
+
+         var startTicks = start.ToUniversalTime().Ticks;
```
ChristoWolf, line 134: I don't know all the ins and outs, but I think it is still [possible to hit problematic transitions using UTC](https://stackoverflow.com/a/40626617).
Could a HPET help with this?

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
+         if (transition.IsFixedDateRule)
+         {
+            startTime = new DateTime(
+               dateTime.Year,
+               transition.Month,
+               transition.Day,
+               transition.TimeOfDay.Hour,
+               transition.TimeOfDay.Minute,
+               transition.TimeOfDay.Second,
+               transition.TimeOfDay.Millisecond);
```
ChristoWolf, line 270: I guess you could reduce nesting by already returning in here.

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
+         }
+         else
+         {
+            var calendar = CultureInfo.CurrentCulture.Calendar;
+
+            var startOfWeek = transition.Week * 7 - 6;
```
ChristoWolf, line 276: Not sure if such `Calendar`s are implemented, but there are real-world calendars which do not follow a 7-day week.

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
+
+            var firstDayOfWeek = (int)calendar.GetDayOfWeek(new DateTime(dateTime.Year, transition.Month, 1));
+            var changeDayOfWeek = (int)transition.DayOfWeek;
+
+            int transitionDay =
```
ChristoWolf, line 281: I guess this could be `var`.

Source/Bogus/DataSets/Date.cs
ChristoWolf: The same is relevant for a few more lines in this method.

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
       {
-         var minTicks = Math.Min(start.Ticks, end.Ticks);
-         var maxTicks = Math.Max(start.Ticks, end.Ticks);
+         var startTime = start.ToUniversalTime().DateTime;
+         var endTime = end.ToUniversalTime().DateTime;
 
-         var totalTimeSpanTicks = maxTicks - minTicks;
+         if (startTime > endTime)
+            return end;
+         else
```
ChristoWolf, line 334: Redundant `else` due to the `return` in the `if` block.

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -1,16 +1,22 @@
...
 using System.Globalization;
+using System.Linq;
 using Bogus.DataSets;
 using FluentAssertions;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Bogus.Tests.DataSetTests
 {
    public class DateTest : SeededTest
```
ChristoWolf, line 11: Just an idea: We could also add a PInvokation to [SetTimeZoneInformation](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-settimezoneinformation) (on Windows) to change the runner's time zone which would allow for covering all time zones e.g. in a  data-driven manner.
This affects system settings though, so I am not sure how compatible this is with the CI runners.

Source/Bogus.Tests/DataSetTests/DateTest.cs
ChristoWolf: I would also prefer consolidation of the test initialization, what's your opinion @bchavez?

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +358,267 @@ public void can_get_timezone_string()
...
+
+         var transitionStartTime = CalculateTransitionDateTime(now, effectiveRule.DaylightTransitionStart);
+
+         // When converting back, .NET picks the end of the transition window instead of the start.
+         var transitionEndTime = transitionStartTime.ToUniversalTime().ToLocalTime();
+
+         // Act
+         var value = faker.Date.Between(transitionStartTime.AddHours(-1), transitionEndTime.AddHours(+2));
+
+         // Assert
```
ChristoWolf, line 395: To prevent defect masking, these could be wrapped by an `AssertionScope`.

ChristoWolf: The same also holds true for other assertion groups in your changes.

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +358,267 @@ public void can_get_timezone_string()
...
+         {
+            var sample = faker.Date.Between(windowStart, windowEnd);
+
+            sample.Should().BeOnOrAfter(transitionEndTime);
+            sample.Should().BeOnOrBefore(windowEnd);
+
+            haveSampleThatIsNotWindowEnd = (sample < windowEnd);
+         }
+
+         haveSampleThatIsNotWindowEnd.Should().BeTrue(because: "the effective range should include values other than windowEnd");
```
ChristoWolf, line 435: I guess this is really nitpicky, so feel free to ignore this:
`windowEnd` in the string would not need to be hard-coded if using `nameof(windowEnd)` instead.

Source/Bogus.Tests/DataSetTests/DateTest.cs
```diff
@@ -352,5 +358,267 @@ public void can_get_timezone_string()
...
+               transition.TimeOfDay.Hour,
+               transition.TimeOfDay.Minute,
+               transition.TimeOfDay.Second,
+               transition.TimeOfDay.Millisecond,
+               DateTimeKind.Local);
+         }
+
+         var calendar = CultureInfo.CurrentCulture.Calendar;
+
+         var startOfWeek = transition.Week * 7 - 6;
```
ChristoWolf, line 600: Again, I still have the question here if `7` is always correct for week-lengths.

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
          var partTimeSpan = RandomTimeSpanFromTicks(totalTimeSpanTicks);
 
-         return new DateTime(minTicks, start.Kind) + partTimeSpan;
+         var value = new DateTime(minTicks, DateTimeKind.Utc) + partTimeSpan;
+
+         if (start.Kind != DateTimeKind.Utc)
+         {
+            value = value.ToLocalTime();
+
+            if (value > end)
```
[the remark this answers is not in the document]
logiclrd: `DateTime` relative comparisons do not factor in the kind, so going from universal to local can change the result of `value > end`. Consider:
```
PS C:\code\Bogus\Source> $a = [DateTime]::new(2023, 3, 19, 11, 0, 0, 0, "Utc")
PS C:\code\Bogus\Source> $b = [DateTime]::new(2023, 3, 19, 11, 0, 0, 0, "Local")
PS C:\code\Bogus\Source> $c = [DateTime]::new(2023, 3, 19, 10, 30, 0, 0, "Local")
PS C:\code\Bogus\Source> $c -lt $b
True
PS C:\code\Bogus\Source> $c -lt $a
True
PS C:\code\Bogus\Source> $c -lt $a.ToLocalTime()
False
PS C:\code\Bogus\Source>
```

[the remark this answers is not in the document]
ChristoWolf: Didn't know that, thanks for the insight!

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
+            {
+               start = new DateTime(window.Start.Ticks, start.Kind);
+
+               if (start == end)
+                  return;
+            }
+
+            window = GetForwardDSTTransitionWindow(end);
+
+            if ((end >= window.Start) && (end < window.End))
```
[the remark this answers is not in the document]
logiclrd: The DST transition window is the span of times that, in the local time zone, _don't exist_ because of the abrupt change do DST. But, `DateTime` comparisons don't take `Kind` into account and just look at the ticks. So, you can create `DateTime` values that point at the start and the end of the window that are _actually the same date/time_, but are numerically different. If you `.ToUtc()` them, you get the same UTC date/time out. This comparison is checking whether the desired end of the range from which to get a bogus DateTime lies within that nebulous region of date/time values that don't actually exist in the real world.

Source/Bogus/DataSets/Date.cs
```diff
@@ -144,31 +129,221 @@ public DateTimeOffset FutureOffset(int yearsToGoForward = 1, DateTimeOffset? ref
...
+               end = new DateTime(window.End.Ticks, end.Kind);
+
+               // We had to bump the end, meaning that the end was not already a valid
+               // DateTime value (within the DST transition range), so prefer the start
+               // instead.
+               preferRangeBoundary = start;
+            }
+
+            if (start > end)
+               throw new Exception("DateTime range does not contain any real DateTime values due to daylight savings transitions");
```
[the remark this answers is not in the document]
logiclrd: If one is using `Faker` to generate `DateTime` values, it will already be avoiding DST windows, so the worst case should be `start == end`. You should never encounter a `start > end` due to DST windows. Or are you talking about potentially pre-existing tests that rely on undefined behaviour by passing in `start` values that are after `end` values and it used to "just work"? This should not be an issue because `ComputeRealRange` swaps out-of-order `start` and `end` before beginning its computations.

Source/Bogus/DataSets/Date.cs
logiclrd: This is a good point, functions that call `SystemClock` multiple times should only call it once and use it as a consistent reference. Otherwise, in extremely rare cases, two consecutive calls will return different values, and the second one _could_ be earlier than the first one. I will make this change.

Source/Bogus/DataSets/Date.cs
logiclrd: I think the visible representation of the branching is a feature, not a bug, personally. :-P Reducing one level of nesting for 20 lines of code comes at a cost of the code not _looking_ like the branch that it is. Also, even though it is only a couple of lines of code in one statement, the `return` statement _is_ doing work and that would need to be duplicated.

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: I am horrified by the idea of unit tests altering system-wide settings. :-P We are between a rock and a hard place on this one. In my opinion, only enabling the tests when the ambient environment supports them is the lesser evil.

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: This sounds like a good thing but it's not something I'm super familiar with. Can you explain in more detail?

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: I am in favour of this kind of thing :-)

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: As far as I can tell, it is always correct _in .NET specifically_. Consider the following documentation, for instance:

> CalendarWeekRule Enum (System.Globalization)
> The first week based on the FirstDay value can have one to seven days. The first week based on the FirstFullWeek value **always has seven days**. The first week ...

([link](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.calendarweekrule?view=net-6.0))

...and the fact that the `DayOfWeek` enum has _exactly_ the western 7 days of the week in it.

Source/Bogus/DataSets/Date.cs
ChristoWolf: Not a bug of course, just a matter of preference :)

Source/Bogus.Tests/DataSetTests/DateTest.cs
ChristoWolf: [This](https://fluentassertions.com/introduction#assertion-scopes) should be good, let me know if I can give you more insight!

Source/Bogus.Tests/DataSetTests/DateTest.cs
ChristoWolf: Good to know, thanks!

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: Ah, nice :-)

logiclrd:
(Somehow I committed/pushed changes that just randomly deleted 7 characters from the middle of a line, force-push makes that commit into what it was supposed to be.

```
diff --git a/Source/Bogus.Tests/DataSetTests/DateTest.cs b/Source/Bogus.Tests/DataSetTests/DateTest.cs
index b8fbdc0..4dc3603 100644
--- a/Source/Bogus.Tests/DataSetTests/DateTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/DateTest.cs
@@ -460,7 +460,7 @@ public void will_adjust_end_time_to_avoid_dst_transition()
          var transitionEndTime = transitionStartTime + effectiveRule.DaylightDelta;

          var windowStart = transitionStartTime.AddMinutes(-30);
-         var nd = transitionStartTime + TimeSpan.FromTicks((transitionEndTime - transitionStartTime).Ticks / 2);
+         var windowEnd = transitionStartTime + TimeSpan.FromTicks((transitionEndTime - transitionStartTime).Ticks / 2);

          // Act & Assert
          using (new AssertionScope())
```
)

logiclrd:
> I would also prefer consolidation of the test initialization, what's your opinion @bchavez?

Making the tests DAMP rather than DRY is a deliberate choice here. Obviously, if those who own the solution have a strong preference then I'll defer to that, but it has repeatedly been my experience that factoring out parts of automated tests almost always results in the tests being harder to grok and harder to maintain.

Source/Bogus.Tests/DataSetTests/DateTest.cs
logiclrd: In my opinion, the correct solution to this is to have the test-running environment prepare virtual machines with desired configuration on which to run the tests. This, of course, doesn't solve the problem for running tests locally, but _that_ I feel _is_ a place where if a developer really needs it, they can temporarily (manually) change their time zone settings.

bchavez:
Quick update:
Finally got the CI/CD build system working again. 😰 PR #471 is now merged.
Next, I'll be going through this PR and @ChristoWolf 's PR review notes as the next to-do item for Bogus soon.

Thanks for the patience.

logiclrd:
How's it looking? :-)

bchavez:
> How's it looking? :-)

Rebased with latest on master; let's see how that goes --

PR #373, opened 2021-05-13
Using switch expression

SergerGood:


bchavez:
Hello,

Thank you for the PR. I appreciate the PR; however, unfortunately, I prefer the older style switch statement especially for the files that are being changed in this PR.

PR #379, opened 2021-06-24
remove redundant Csproj Stuff

SimonCropp:


Source/Bogus/Bogus.csproj
```diff
@@ -8,11 +8,6 @@
     <TargetFrameworks>net40;netstandard1.3;netstandard2.0</TargetFrameworks>
     <CodeAnalysisRuleSet>Bogus.ruleset</CodeAnalysisRuleSet>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <AssemblyOriginatorKeyFile>
-    </AssemblyOriginatorKeyFile>
-    <SignAssembly>false</SignAssembly>
-    <AssemblyName>Bogus</AssemblyName>
-    <PackageId>Bogus</PackageId>
```
bchavez, line 15: Hey @SimonCropp, most of these attributes in `*.csproj` are expected to be present in `*.csproj` due to the F# FAKE build script mutating the `*.csproj` files during the build & release process.

https://github.com/bchavez/Bogus/blob/9fc7614274f7f80b889151aa31a5fd2f2d0e610d/Source/Builder/build.fsx#L245-L251

If these attributes are missing from `*.csproj`, this would essentially break the build because the F# FAKE `build.fsx` file is expecting them to be present.

I think this is why the CI/CD build broke with this PR:

image

Ideally, it would be great if we can keep these specific attributes present in the `*.csproj`. 👍 So, I'm not sure we would merge this PR.

SimonCropp:
sorry for the noise

PR #390, opened 2021-08-13
Fix setting property with private setter on base type

### LINKED ISSUE #389: Properties with private setters are not assigned if the property is declared on the base class
### Version Information
| Software                       | Version(s) |
| ------------------------| ---------- |
| Bogus NuGet Package  | 32.0.2 |
| .NET Core?                   | .NET 5.0 |
| .NET Full Framework?   | N/A |
| Windows OS?              | 10|
| Linux OS?                    | N/A |
| Visual Studio?              | irrelevant (I use Rider) |

en-US (probably irrelevant)

### What is the expected behavior?

```csharp
public class FooBase
{
    public int Baz { get; private set; }
}

public class Foo : FooBase
{
    public int Bar { get; private set; }
}

var foo = new Faker<Foo>()
        .RuleFor(f => f.Bar, 42)
        .RuleFor(f => f.Baz, 123)
        .Generate();
```

foo.Bar should be 42, foo.Baz should be 123

### What is the actual behavior?

foo.Bar is 42, but foo.Baz is 0

The only difference I can see between the two is that Baz is declared in the base class.
It works fine if I make Baz's setter public.

### Please provide a stack trace.

N/A (no exception)

### Any possible solutions?

Make the setter public (not acceptable in my case)

### How do you reproduce the issue?

See code above

### Do you have a unit test that can demonstrate the bug?

No, but easy to write using the code above

### Can you identify the location in Bogus' source code where the problem exists?

Not yet

### If the bug is confirmed, would you be willing to submit a PR?

Yes

thomaslevesque:
Fixes #389 

Not very pretty, but doesn't seem to break anything...

### DIFF
--- a/Source/Bogus/Binder.cs
+++ b/Source/Bogus/Binder.cs
@@ -61,6 +61,7 @@ namespace Bogus
       public virtual Dictionary<string, MemberInfo> GetMembers(Type t)
       {
          var group = t.GetAllMembers(BindingFlags)
+            .Select(m => GetPropertyOnBaseType(m, t))
             .Where(m =>
                {
                   if( m.GetCustomAttributes(typeof(CompilerGeneratedAttribute), true).Any() )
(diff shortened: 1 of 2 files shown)


Source/Bogus/Binder.cs
```diff
@@ -94,5 +95,18 @@ public Binder(BindingFlags bindingFlags)
          //ClassA.Value.
          return group.ToDictionary(k => k.Key, g => g.First());
       }
+
+      private static MemberInfo GetPropertyOnBaseType(MemberInfo member, Type reflectedType)
```
thomaslevesque, line 99: Note: I passed the reflected type here, because the `ReflectedType` property isn't available in netstandard1.3.

PR #396, opened 2021-10-18
even end digit for females, odd for males

PTST:
According to the CPR register i Denmark, all female CPR numbers should end with an even number and odd numbers for male


bchavez:
I think the CPR number generator was recently improved in PR #496 takes into account the Male:Odd/Female:Evenness and should be in Bogus `v35.3.0+`.

I think we can close this PR. Thanks again for the initial PR @PTST.

Issue #411, opened 2022-02-14
Allow for different Pseudo Random Number Generators (PRNGs)


DaveRMaltby:
### Please describe why you are requesting a feature
Different [PRNGs](https://www.geeksforgeeks.org/pseudo-random-number-generator-prng/) have different tradeoffs on CPU, memory and distribution of results.  Bogus is currently bound to using the [.NET Random class](https://docs.microsoft.com/en-us/dotnet/api/system.random?view=net-6.0) which is constrained to only one PRNG algorithm (See remarks of the previous link).  Bogus is designed in such a manner that adding this feature requires minimal changes.

### Please provide a code example of what you are trying to achieve

```
namespace Bogus
{
   public interface IRandom
   {
      int Next();
      int Next(int minValue, int maxValue);
      void NextBytes(byte[] buffer);
      double NextDouble();
   }

   internal class RandomWrapper : IRandom
   {
      private readonly Random random;

      public RandomWrapper(Random random)
      {
         this.random = random ?? throw new ArgumentNullException(nameof(random));
      }

      public int Next()
      {
         return random.Next();
      }

      public int Next(int minValue, int maxValue)
      {
         return random.Next(minValue, maxValue);
      }

      public void NextBytes(byte[] buffer)
      {
         random.NextBytes(buffer);
      }

      public double NextDouble()
      {
         return random.NextDouble();
      }
   }

   public class Randomizer
   {
...

//CHANGED from the data type Random
      private readonly IRandom localSeed;

      public Randomizer()
      {
//CHANGED
         this.localSeed = new RandomWrapper(Seed);
      }

      public Randomizer(int localSeed)
      {
//CHANGED
         this.localSeed = new RandomWrapper(new Random(localSeed));
      }

...
      /// <summary>
      /// Constructor that uses <see cref="random"/> parameter interface as an abstraction for Random.
      /// Completely ignores the global static <see cref="Seed"/>.
      /// </summary>
      public Randomizer(IRandom random)
      {
         this.localSeed = random;
      }
...
   }
}

namespace Bogus.Tests
{
      private class PseudoRandomNumberGenerator : IRandom
      {
         public int Next() => (int)(DateTime.Now.Ticks & 0xFFFFFFFFL);
         public int Next(int minValue, int maxValue) => Next() % (maxValue - minValue) + minValue;
         public void NextBytes(byte[] buffer) => throw new NotImplementedException();
         public double NextDouble() => throw new NotImplementedException();
      }

      [Fact]
[...]

PR #412, opened 2022-02-19
#411 - Allow for different PRNGs (random algorithms)

DaveRMaltby:
Provides a new IRandom interface that is a sub-set of the methods of the System.Random class.

### DIFF
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -562,7 +562,7 @@ namespace Bogus.Tests
             .Where(ex => ex.Message.StartsWith("The collection is empty. There are no items to select."));
       }
 
-      private class PseudoRandomNumberGenerator : IRandom
+      private class CustomRandom : IRandom
       {
          public int Next() => (int)(DateTime.Now.Ticks & 0x7FFFFFFFL);
          public int Next(int minValue, int maxValue) => Next() % (maxValue - minValue) + minValue;
@@ -574,7 +574,7 @@ namespace Bogus.Tests
       public void custom_random_interface()
       {
          var nameAlg = new DataSets.Name();
-         nameAlg.Random = new Randomizer(new PseudoRandomNumberGenerator());
+         nameAlg.Random = new Randomizer(new CustomRandom());
 
          string firstName = string.Empty;
          var exception = Record.Exception(() => firstName = nameAlg.FirstName());
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -561,5 +561,26 @@ namespace Bogus.Tests
          collectionAction.Should().Throw<ArgumentException>()
             .Where(ex => ex.Message.StartsWith("The collection is empty. There are no items to select."));
       }
+
+      private class PseudoRandomNumberGenerator : IRandom
+      {
+         public int Next() => (int)(DateTime.Now.Ticks & 0xFFFFFFFFL);
+         public int Next(int minValue, int maxValue) => Next() % (maxValue - minValue) + minValue;
+         public void NextBytes(byte[] buffer) => throw new NotImplementedException();
+         public double NextDouble() => throw new NotImplementedException();
+      }
+
+      [Fact]
+      public void custom_random_interface()
+      {
+         var nameAlg = new DataSets.Name();
+         nameAlg.Random = new Randomizer(new PseudoRandomNumberGenerator());
+
+         string firstName = string.Empty;
+         var exception = Record.Exception(() => firstName = nameAlg.FirstName());
+
+         Assert.Null(exception);
+         Assert.False(string.IsNullOrEmpty(firstName));
+      }
    }
 }
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -564,7 +564,7 @@ namespace Bogus.Tests
 
       private class PseudoRandomNumberGenerator : IRandom
       {
-         public int Next() => (int)(DateTime.Now.Ticks & 0xFFFFFFFFL);
+         public int Next() => (int)(DateTime.Now.Ticks & 0x7FFFFFFFL);
          public int Next(int minValue, int maxValue) => Next() % (maxValue - minValue) + minValue;
          public void NextBytes(byte[] buffer) => throw new NotImplementedException();
          public double NextDouble() => throw new NotImplementedException();
--- a/Source/Bogus/IRandom.cs
+++ b/Source/Bogus/IRandom.cs
@@ -1,0 +1,62 @@
+﻿
+namespace Bogus
+{
+   /// <summary>
+   /// An interface that abstracts the subset of System.Random that is used by Bogus.
+   /// </summary>
+   public interface IRandom
+   {
+      //
+      // Summary:
+      //     Returns a non-negative random integer.
+      //
+      // Returns:
+      //     A 32-bit signed integer that is greater than or equal to 0 and less than System.Int32.MaxValue.
+      int Next();
+
+      //
+      // Summary:
+      //     Returns a random integer that is within a specified range.
+      //
+      // Parameters:
+      //   minValue:
+      //     The inclusive lower bound of the random number returned.
+      //
+      //   maxValue:
+      //     The exclusive upper bound of the random number returned. maxValue must be greater
+      //     than or equal to minValue.
+      //
+      // Returns:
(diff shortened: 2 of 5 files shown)


Source/Bogus/IRandom.cs
```diff
@@ -0,0 +1,62 @@
...
+      // Summary:
+      //     Returns a random integer that is within a specified range.
+      //
+      // Parameters:
+      //   minValue:
+      //     The inclusive lower bound of the random number returned.
+      //
+      //   maxValue:
+      //     The exclusive upper bound of the random number returned. maxValue must be greater
+      //     than or equal to minValue.
```
bchavez, line 27: You will need to use proper C# XMLdoc markup.

Source/Bogus/RandomWrapper.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+
+namespace Bogus
+{
+   /// <summary>
+   /// A wrapper around System.Random that supports the IRandom interface.
+   /// </summary>
+   internal class RandomWrapper : IRandom
```
bchavez, line 8: Rename this to `DefaultRandom`

Source/Bogus.Tests/RandomizerTest.cs
```diff
@@ -561,5 +561,26 @@ public void empty_collection_throws_better_exception_message_rather_than_index_o
...
+      {
+         var nameAlg = new DataSets.Name();
+         nameAlg.Random = new Randomizer(new PseudoRandomNumberGenerator());
+
+         string firstName = string.Empty;
+         var exception = Record.Exception(() => firstName = nameAlg.FirstName());
+
+         Assert.Null(exception);
+         Assert.False(string.IsNullOrEmpty(firstName));
+      }
```
bchavez, line 584: There is a problem with this test because it failed CI/CD. You'll need to investigate the failure. Check details in the AppVeyor GH status checks.

bchavez (changes requested):

Source/Bogus.Tests/RandomizerTest.cs
```diff
@@ -561,5 +561,26 @@ public void empty_collection_throws_better_exception_message_rather_than_index_o
          collectionAction.Should().Throw<ArgumentException>()
             .Where(ex => ex.Message.StartsWith("The collection is empty. There are no items to select."));
       }
+
+      private class PseudoRandomNumberGenerator : IRandom
```
bchavez, line 565: Additionally, rename this to `CustomRandom`.

bchavez:
Thank you @DaveRMaltby for the PR; I think this PR is ready for review now. I'm pretty sure I can take over from here and carry it across the finishing line.

I will need some time to think about this PR to make sure it's a good fit; I have to weigh the pros/cons, cost/benefits, and value.

I know this PR might be of high value for you; but when I review this PR later, I need to get a better understanding of:
* How valuable is this PR to others? How many others are going to need a different PRNG other than the one .NET provides?
* What kind of new support/maintenance issues am I going to have to serve because of this change?
* Is it worth the extra abstraction layer? Or does some other higher-level architectural change achieve a better result in terms of keeping things simple and consistent?

Thanks,
Brian

bchavez:
Review notes:
* Thinking about making a breaking change for `Randomizer.Seed` to use `IRandom`; or at least `[Obsolte]` the usage.
* Make `DefaultRandom : System.Random, IRandom`; so `new DefaultRandom()` is just like `new System.Random()`
* APIs exposing `System.Random` and `IRandom` leads to confusion on the same object

DaveRMaltby:
Brian,
I agree with your assessment here.  As the repo owner, I totally understand the need to weigh these aspects.  I appreciate you considering my need and thank you for even considering it.  I like your suggestion of having DefaultRandom inherit Random.  I assumed that System.Random would be sealed, but now see that it is not.

Regards,
David

bchavez:
I had the **_same_** exact assumption - I originally thought `System.Random` was _sealed_ too. It seemed more natural to subclass and tac on `IRandom` for this particular requirement; then confirmed via decompilation and was happily surprised random was not _sealed_. I'm not a big fan of wrapper classes :grimacing: so I was just experimenting.

But when I pursued that path; a bunch of tests broke because they assumed `Randomizer.Seed<System.Random>` and I changed the contract to `Randomizer.Seed<IRandom>`; and I was like... oh yeah... so, I know have to weigh that too... I'm okay with breaking changes generally if there's enough justification for it and it produces better clarity and maintainability.

----

Out of curiosity, have you tried to subclass `System.Random` and override method signatures defined in `IRandom`?

I just now checked the **.NET Full Framework** and **.NET Core** sources, and it seems that `System.Next*()` are marked `virtual`. Do you think you could get away with a custom PRNG implementation without this PR?

DaveRMaltby:
Clever thinking.  I'll have to check on that.  No, I haven't tried subclassing System.Random.  I'll try that.  My only hesitation there, is that I was trying to avoid the expense of the Random ctor.  In my previous testing, the seeded ctor was as expensive as calling Next() about 1.2 million times..  but maybe I can make something work that doesn't kill performance in my situation.  I'm potentially constructing thousands of PRNGs here.  Thanks again.

bchavez:
The following seems to work:

```csharp
void Main()
{
   Randomizer.Seed = new CustomRandom();
   
   var r = new Randomizer();
   r.Number().Dump();
   r.Number().Dump();
   r.Number().Dump();
   r.Number().Dump();
   r.Number().Dump();
   r.Number().Dump();
}


public class CustomRandom : System.Random
{
   public override int Next()
   {
      return 1;
   }
   public override int Next(int maxValue)
   {
      return 1;
   }
   public override float NextSingle()
   {
      return 1.0f;
   }
   public override long NextInt64()
   {
      return 1;
   }
   public override void NextBytes(byte[] buffer)
   {
      buffer[0] = 1;
   }
   public override int Next(int minValue, int maxValue)
   {
      return 1;
   }
   public override long NextInt64(long maxValue)
   {
      return 1;
   }
   public override double NextDouble()
   {
      return 1.0;
   }
   public override void NextBytes(Span<byte> buffer)
   {
      base.NextBytes(buffer);
   }
   public override long NextInt64(long minValue, long maxValue)
   {
      return 1;
   }
}
```
```
Output:
1
1
1
1
1
1
```

The only disadvantage right now would be setting a `localSeed` for `Randomizer` by `System.Random` instance; which could be just an easy change as opening up a new constructor overload for `new Rnadomizer(System.Random)`.

bchavez:
Ah, okay, cool. Let me know what you find.

DaveRMaltby:
Yes, this approach of inheriting from `Random` would suffice for me.  I would only need the new ctor overload that you mentioned of `new Randomizer(System.Random)`... In my work, I'm constructing so many different `Randomizer` instances that the `public static Random Seed` would be hard to use.

DaveRMaltby:
Please let me know if you'd like for me to submit a PR for just the added `Randomizer` ctor that takes a `System.Random`.  Thanks!

bchavez:
@DaveRMaltby I think there are two possibilities:

1. Create a `new Randomizer(System.Random)` constructor. 
2. Also, we can make `localSeed` `protected`; so that it allows inheriting of `Randomizer` with the ability to set your own `localSeed` with a custom constructor; eg:
   ```csharp
      public class CustomRandomizer : Randomizer
      {
         public CustomRandomizer(System.Random myRandom)
         {
            this.localSeed = myRandom;
         }
      }
   ```
   https://github.com/bchavez/Bogus/blob/59985ee11fb2a34b89d940d5da8b34617f6f9f6f/Source/Bogus/Randomizer.cs#L44

IMHO, I think both of the options above are better than our `IRandom` abstraction layer.

I'm 50/50; but slightly leaning on No.2 if that one works for you... - typically, in situations like this; I've opened up `internal` or `private` members - usually, it keeps public "surface" APIs simple and ergonomic, and for those situations that need to take the extension to the next level - the `protected` non-public members are usually a way to wire-in extensibility or different behavior under the hood.

If you're up for it, you can close this PR and begin a different approach; just as always, please be sure to include unit tests; I rely heavily on the unit tests as a "go/no-go/release-ready/breaking-change/software-quality" signal. Additionally, any new unit tests should be tied to the github issue that raised the change so I can easily remember the context about why that change was needed since this change was initiated by a customer/user. Eg `Issue411.cs`:
* https://github.com/bchavez/Bogus/tree/master/Source/Bogus.Tests/GitHubIssues

DaveRMaltby:
@bchavez ,
#2 it is.  Both options work for me.  I've got some important business this evening, I'll close this PR now and tomorrow create a new one with unit tests in an `Issue411.cs` file.  Appreciate again, you working to accommodate me here!

PR #413, opened 2022-02-22
#411 - Change Randomizer.localSeed to be protected, so that derived classes …

DaveRMaltby:
…can inject their own PRNG.

### DIFF
--- a/Source/Bogus/Randomizer.cs
+++ b/Source/Bogus/Randomizer.cs
@@ -41,7 +41,10 @@ namespace Bogus
          this.localSeed = new Random(localSeed);
       }
 
-      private readonly Random localSeed;
+      /// <summary>
+      /// The pseudo-random number generator that is used for all random number generation in this instance.
+      /// </summary>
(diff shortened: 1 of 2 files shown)


Source/Bogus/Randomizer.cs
```diff
@@ -41,7 +41,10 @@ public Randomizer(int localSeed)
          this.localSeed = new Random(localSeed);
       }
 
-      private readonly Random localSeed;
+      /// <summary>
+      /// The pseudo-random number generator that is used for all random number generation in this instance.
+      /// </summary>
+      protected Random localSeed;
```
DaveRMaltby, line 47: It surprised me that I couldn't keep `localSeed` as readonly.  But from what I'm reading in the C# specification, the readonly applies only to the ctor of the class where it is declared.  (i.e. it can't be set in the ctor of derived classes)

bchavez: No problemo. That should be fine.

PR #415, opened 2022-03-12
Fix docstring that used `example` tag instead of `exception`

DanteDeRuwe:
Probable typo. 

This makes sure that hovering over the method in IDE's has the relevant display for these exceptions. With the example tag, not even the name of the exception that was going to throw was visible.


DanteDeRuwe:
Can't seem to get rid of the file ending diff 😞

Source/Bogus/Faker.cs
```diff
@@ -383,4 +383,4 @@ internal void NewContext()
       /// </summary>
       public Hashids Hashids { get; set; }
    }
-}
```
DanteDeRuwe, line 386: Can't seem to get rid of the file ending diff 😞

PR #421, opened 2022-04-22
Add option to generate strict VIN (exclusive characters I, O and Q)

### LINKED ISSUE #449: Bogus generates invalid VINs. VIN can't contain letters I, O and Q
https://eu.statesman.com/story/news/2012/09/25/what-the-vin-tells-you-about-your-vehicle/9848258007/

According to federal law, the VIN must appear on a part of the vehicle that won’t be removed. Each character in the VIN must be either a letter from the alphabet or a number from 0 through 9. They cannot contain the letters I, O or Q to help cut down on mistaken characters. VINs appear on the left side of the dash and must be visible through the windshield. The VIN is also stamped on the frame and other locations in case of a fire.

https://github.com/bchavez/Bogus/blob/b9049abf8b40203c09079741bcb328da95899f81/Source/Bogus/DataSets/Vehicle.cs#L13-L23

skwasjer:
Adds an optional boolean to opt-in to generating a VIN using a more strict characterset.

> In 1981, the [National Highway Traffic Safety Administration](https://en.wikipedia.org/wiki/National_Highway_Traffic_Safety_Administration) of the United States standardized the format.[[1]](https://en.wikipedia.org/wiki/Vehicle_identification_number#cite_note-nhtsa1-1) It required all on-road vehicles sold to contain a 17-character VIN, which does not include the letters O (o), I (i), and Q (q) (to avoid confusion with numerals 0, 1, and 9).

Fixes #449

### DIFF
--- a/Source/Bogus.Tests/DataSetTests/VehicleTest.cs
+++ b/Source/Bogus.Tests/DataSetTests/VehicleTest.cs
@@ -27,6 +27,12 @@ public class VehicleTest : SeededTest
    }
 
    [Fact]
+   public void can_get_a_strict_vin_number()
+   {
+      vehicle.Vin(true).Should().Be("K3TM1L1NF9Y575714");
+   }
+
+   [Fact]
    public void can_get_a_manufacture()
    {
       vehicle.Manufacturer().Should().Be("Maserati");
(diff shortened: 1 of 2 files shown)


skwasjer:
Any ETA on this? :)

ps: I could also remove the bool argument and just make this the default behavior if so desired.

There's also a bug I found whilst working on this, see #431

bchavez:

Thanks again for the help.
-Brian

PR #427, opened 2022-06-12
Fix incorrect generation for Finnish social security number

### LINKED ISSUE #392: Bogus generates invalid Finnish Henkilötunnus
### Version Information
| Software                       | Version(s) |
| ------------------------| ---------- |
| Bogus NuGet Package  |       33.1.1          |
| .NET Core?                   |      3.1           |
| .NET Full Framework?   |                 |
| Windows OS?              |     N/A        |
| Linux OS?                    |           N/A        |
| Visual Studio?              |       N/A           |

Default, en

### What is the expected behavior?
API Extension method Bogus.Person.Henkilötunnus() to return a valid Henkilötunnus.

### What is the actual behavior?
I believe the algorithm for calculating the control digit is wrong.

### Please provide a stack trace.
N/A

### Any possible solutions?
Implement the correct algorithm for the control digit: https://en.wikipedia.org/wiki/National_identification_number#Finland

### How do you reproduce the issue?
```
   var person = new Faker<Person>()
                .RuleFor(p => p.DateOfBirth, f => DateTime.Parse("1989-07-14"))
                .CustomInstantiator(f => new Person())
                .Generate();
            
            var hetu = person.Henkilötunnus();
```
will generate a control digit of I (upper case i) which is not an allowed character

### Do you have a unit test that can demonstrate the bug?
```
        [Fact]
        public void ShouldProduceValidHetu()
        {
            var person = new Person();
            var hetu = person.Henkilötunnus();
            var expectedControlDigit = CalculateControlDigit(hetu);
            var actualControlDigit = hetu[^1];
            Assert.Equal(expectedControlDigit, actualControlDigit);
        }

        private char CalculateControlDigit(string hetu)
        {
            const string allowedCharacters = "0123456789ABCDEFHJKLMNPRSTUVWXY";
            var removedDelimiters = hetu
                .Replace("-", string.Empty)
                .Replace("+", string.Empty)
                .Replace("A", string.Empty);
                
            var ddMMyyzzz =  
                removedDelimiters
                .Remove(removedDelimiters.Length - 1);
            var n = int.Parse(ddMMyyzzz) % 31;
            return allowedCharacters[n];
        }
```

### Can you identify the location in Bogus' source code where the problem exists?
Bogus.Extensions.Finland.Henkilötunnus()
```
         var ddMMyy = $"{p.DateOfBirth:ddMMyy}";

         var n = int.Parse(ddMMyy) % 31;

         var q = n.ToString();
         if( n >= 10 )
            q = ((char)('A' + (n - 10))).ToString();

         //no idea if its female or male.
         var zzz = r.Replace("###");
```
zzz is created after the control digit has been set, which is wrong because the control digit depends on the zzz value.

### If the bug is confirmed, would you be willing to submit a PR?

Yes, I wrote an implementation i think works above.

roeyskoe:
There were three issues with calculating control character for Finnish personal identity code (henkilötunnus).
1. Three-digit part after birthdate was not taken into account when calculating it.
2. Invalid characters were not skipped.
3. Gender was ignored.

It was also (rarely) possible to get 000 or 001 as the number part, both of which invalid.

This also changes that the number part is always between 002-899. Numbers 900-999 do not exist "in the wild", they are strictly reserved for temporary identity codes.
There could of course be an option to also allow temporary codes, or generate only them. I can add them if you wish.

How the identity code is formed: https://dvv.fi/en/personal-identity-code

Fixes #392

### DIFF
--- a/Source/Bogus/Extensions/Finland/ExtensionsForFinland.cs
+++ b/Source/Bogus/Extensions/Finland/ExtensionsForFinland.cs
@@ -34,7 +34,7 @@ public static class ExtensionsForFinland
       // gives the remainder as 30, and since A=10, B=11, etc. ending up with Y=30.
       // Note that letters G, I, O, and Q are skipped.
 
-      var controlCharTable = "0123456789ABCDEFHJKLMNPRSTUVWXY";
+      const string controlCharTable = "0123456789ABCDEFHJKLMNPRSTUVWXY";
       var r = p.Random;
 
       var year = p.DateOfBirth.Year;
--- a/Source/Bogus/Extensions/Finland/ExtensionsForFinland.cs
+++ b/Source/Bogus/Extensions/Finland/ExtensionsForFinland.cs
@@ -1,4 +1,6 @@
-﻿namespace Bogus.Extensions.Finland;
+﻿using Bogus.DataSets;
+
+namespace Bogus.Extensions.Finland;
 
 /// <summary>
 /// API extensions specific for a geographical location.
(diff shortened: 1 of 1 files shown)


bchavez:

* Changelog: https://github.com/bchavez/Bogus/blob/master/HISTORY.md

Thanks again for the help.
-Brian

PR #431, opened 2022-06-25
Fixes VIN sometimes has 18 character length

skwasjer:
The last appended random number segment of a VIN could sometimes yield an invalid VIN because the max allowed number value was 100000. This caused the returned VIN to have a length of 18 characters.

Without the fix, the new test case in this PR would produce an assertion exception:
```
Expected string with length 17, but found string "XTVJ5JFU2YBV100000" with length 18.
```


bchavez:

* Changelog: https://github.com/bchavez/Bogus/blob/master/HISTORY.md

Thanks again for the help.
-Brian

PR #461, opened 2023-02-01
Method to return numbers only string with a specific length...

nirzaf:
Method to return a numbers-only string with a specific length defined via parameter

### DIFF
--- a/Source/Bogus.Tests/RandomizerTest.cs
+++ b/Source/Bogus.Tests/RandomizerTest.cs
@@ -374,6 +374,12 @@ namespace Bogus.Tests
       {
          r.String2(5).Should().Be("pcvqa");
       }
+      
+      [Fact]
+      public void generate_numbers_only_pool()
+      {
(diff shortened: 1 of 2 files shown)


ChristoWolf:
Hi @nirzaf!

This is easily achievable using [any of the `Replace` methods](https://github.com/bchavez/Bogus/blob/c52e76f9fb30047578b83f723448f072c7d3670b/Source/Bogus/Randomizer.cs#L605), so this proposal would unnecessarily bloat the API IMO

Is there anything that you are missing from the current API?

bchavez:
Hi. Thank you for the feedback.
I think @ChristoWolf is right on this one for the moment. You'll need to create a custom extension method for this if you want better API ergonomics. If there's enough demand for this new method to be in the Bogus API, then we'll consider it in the future; but for now, your best advice is to create a custom extension method; or use the `Random.Replace*()` methods as shown below:

```csharp
void Main()
{
   var f = new Faker();

   f.Random.ReplaceNumbers(new string('#', 25)).Dump();
   f.Random.NumberOnlyString(25).Dump();;
}

public static class MyCustomExtensions
{
   public static string NumberOnlyString(this Randomizer r, int length)
   {
      return r.ReplaceNumbers(new string('#', length));
   }
}
//OUTPUT:
9618188693116119676746702
3035736940236164577423885
```

PR #483, opened 2023-07-13
United Kingdom Extensions: Added NHSNumber generator to Person

JohnStabler:
NHS numbers at 10 digits long. The last digit is a check digit


bchavez:
Hi. I think we need a unit test for this. Also, is there a way to generate an NHS number without the `do/while` loop?

Issue #474, opened 2023-04-27
Adding Chilean Bogus RUT


acidrod:
### Please provide a code example of what you are trying to achieve

```
public static class CustomBogusExtensions
{
    /// <summary>
    /// Rol Unico Tributario Chileno
    /// </summary>
    /// <param name="DotFormat"></param>

    public static string Rut(this Person person, bool DotFormat = true)
    {
        Random rnd = new();
        int num = rnd.Next(1000000, 99999999);
        string dig = Digito(num);

        if (DotFormat)
        {
            var rut = string.Format("{0:0,0}", num);
            return $"{rut}-{dig}";
        }
        
        return $"{num}-{dig}";
    }

    private static string Digito(int rut)
    {
        int suma = 0;
        int multiplicador = 1;

        while (rut != 0)
        {
            multiplicador++;
            if (multiplicador == 8)
                multiplicador = 2;
            suma += (rut % 10) * multiplicador;
            rut /= 10;
        }

        suma = 11 - (suma % 11);

        if (suma == 11)
        {
            return "0";
        }
        else if (suma == 10)
        {
            return "K";
        }
        else
        {
            return suma.ToString();
        }
    }
}
```

### Please answer any or all of the questions below

* Is the feature something that currently cannot be done?
Yes, this cannot be done since it's a missing feature

* What alternatives have you considered?
I already created my own Extension Method and it works very well.

* Has the feature been requested in the past?
No, I searched on the issues and it does not appear to be requested before

**If the feature request is approved, would you be willing to submit a PR?**

Yes  _(Help can be provided if you need assistance submitting a PR)_

PR #487, opened 2023-07-30
Added new Chilean RUT generator

acidrod:
Created a full Chilean RUT generator for extending the Person and Company classes (Both of them use the Rut ID). The tests are fully validating the generation for random IDs.

### DIFF
--- a/Source/Bogus/Extensions/Chile/ExtensionsForChile.cs
+++ b/Source/Bogus/Extensions/Chile/ExtensionsForChile.cs
@@ -1,0 +1,92 @@
+﻿using System;
+using Bogus.DataSets;
+
+namespace Bogus.Extensions.Chile
+{
+    /// <summary>
+    /// API extensions specific for a geographical location.
+    /// </summary>
+    public static class ExtensionsForChile
+    {
+      /// <summary>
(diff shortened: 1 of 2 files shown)


acidrod:
This PR addresses the Issue #474

Source/Bogus/Extensions/Chile/ExtensionsForChile.cs
```diff
@@ -0,0 +1,92 @@
...
+
+      /// <summary>
+      /// A general Chilean ID generator
+      /// </summary>
+      /// <param name="dotFormat">Use the thousands separator for the Chilean RUT (11.111.111-1)</param>
+      /// <returns></returns>
+      private static string GenerateChileanId(bool dotFormat = true)
+      {
+         Random rnd = new();
+         int num = rnd.Next(1000000, 99999999);
```
bchavez, line 40: We shouldn't call `new Random()` in Bogus because it's not deterministic. We'll need to source the `rnd` from `Person.Random` to get the randomizer, and then ask for the `rnd.Next()`.

bchavez (changes requested):

PR #496, opened 2023-09-04
Denmark extensions: Updates CPR number generator with checksum

MunroRaymaker:
Some older systems still use CPR number checksum validation. This updates the danish CPR number extension, to optionally create CPR numbers that can be validated against checksums.
Unit tests are also provided for the extension updates.

### DIFF
--- a/Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
+++ b/Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
@@ -1,4 +1,7 @@
-﻿namespace Bogus.Extensions.Denmark;
+﻿using static Bogus.DataSets.Name;
+using System;
+
+namespace Bogus.Extensions.Denmark;
 
 /// <summary>
 /// API extensions specific for a geographical location.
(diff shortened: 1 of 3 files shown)


Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
```diff
@@ -8,19 +11,169 @@ public static class ExtensionsForDenmark
...
+               hasValidChecksum = GenerateChecksum(birthDate, p.Gender, individualNumber, out checksum);
+            } while (!hasValidChecksum);
+         }
+         else
+         {
+            checksum = string.Empty;
+            individualNumber = GenerateIndividualFourDigitNumber(r, p.Gender, p.DateOfBirth.Year);
+         }
+
+         var final = $"{birthDate}-{individualNumber}{checksum}";
```
NielsPilgaard, line 67: It would be nice to have a parameter to toggle this dash off.

NielsPilgaard (approved):
I find this change very helpful, and the logic looks good. I'm not a Bogus Contributor though, so I'm not sure my approval helps much 😅 

The current CPR implementation in Bogus doesn't generate valid CPR numbers, which is a bit counter-intuitive to me.

MunroRaymaker:
Struggling to push a commit that resolve the conflicts. A lot has happened in master since I created this pullrequest.
I getting this error when I try a git push:

> JRN@DK-PC5CG3070Q8R MINGW64 /c/src/github/Bogus_branch_with_changes_to_danish_cpr/Source (dk-extensions-cpr-number)
> $ git push
> fatal: The current branch dk-extensions-cpr-number has no upstream branch.
> To push the current branch and set the remote as upstream, use
> 
>     git push --set-upstream origin dk-extensions-cpr-number
> 
> To have this happen automatically for branches without a tracking
> upstream, see 'push.autoSetupRemote' in 'git help config'.
> 

Don't understand why I need to create an upstream branch. Thought that was already there.

NielsPilgaard:
It does look like the upstream branch is present on your repo. Perhaps it would be easiest to start over on a fresh branch?

bchavez:
Merge conflict should be resolved;
The crux of the problem is a conversion to file-scoped namespaces;
If the build passes; we will merge. Thanks again for the PR.

bchavez:
@NielsPilgaard @MunroRaymaker if you could have one last look to ensure the rebase was successful that would be great.
Lots of thanks,
Brian

Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
```diff
@@ -8,18 +11,168 @@ public static class ExtensionsForDenmark
...
+         to = 9999;
+      }
+      else if (2037 <= year && year <= 2057)
+      {
+         from = 5000;
+         to = 9999;
+      }
+      else
+      {
+         throw new ArgumentOutOfRangeException(nameof(year), $"{nameof(year)} must be between 1854 and 2039.");
```
NielsPilgaard, line 105: The else if above suggest the range ends at 2057, which year is correct?

Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
```diff
@@ -8,18 +11,168 @@ public static class ExtensionsForDenmark
...
+         to = 999;
+      }
+      else if (2037 <= year && year <= 2057)
+      {
+         from = 500;
+         to = 999;
+      }
+      else
+      {
+         throw new ArgumentOutOfRangeException(nameof(year), $"{nameof(year)} must be between 1854 and 2039.");
```
NielsPilgaard, line 145: The else if above suggest the range ends at 2057, which year is correct?

NielsPilgaard (approved):
Looks good to me still, everything is intact after the merge. 
I spotted and commented on 2 minor issues, but they're not blocking in my opinion.

Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
bchavez: The `includeDash` parameter has been added in the latest commit

Source/Bogus/Extensions/Denmark/ExtensionsForDenmark.cs
bchavez: This should be fixed in the latest commit. Please have a look.

NielsPilgaard (approved):
LGTM, thanks for adding the include dash parameter 👍

MunroRaymaker:
Thanks guys for help with the PR 🥇

Issue #528, opened 2024-02-09
Are Randomizer's number ranges inclusive or exclusive?


Mitchman215:
### Bogus NuGet Package

34.0.2

### .NET Version

.NET 8

### What operating system are you using?

Windows

### What locale are you using with Bogus?

en_US

### What's the problem?

It doesn't seem to be clearly documented whether the number ranges in `Faker.Random` are inclusive or exclusive. For example, `Random.Double`'s documentation reads "Get a random double, between 0.0 and 1.0."

### Provide LINQPad Example

NA

### What other possible solutions or alternatives have you considered?

Looked at README docs and individual function docs

bchavez:
You can also check the source code:

https://github.com/bchavez/Bogus/blob/1774a851db082b5071291620f35bf8d53fd4dc15/Source/Bogus/Randomizer.cs#L173-L191

Looks like in this specific case, `Randomizer.Double()` defers to `System.Random.NextDouble()`:
* https://learn.microsoft.com/en-us/dotnet/api/system.random.nextdouble?view=net-8.0

Mitchman215:
Thanks for the quick answer! Would you be opposed to me opening up a PR adding this documentation, just so it's a bit easier to determine?

bchavez:
@Mitchman215 , of course, go for it.

Just a couple of guidelines:
* Reference this GH issue number `#528` in your PR.
* Also, feel free to make the **_same kind_** of doc improvements (inclusive/exclusive) in `Randomizer.cs` if they are missing in other XML param docs for other methods.
  * But please keep your PR/change-set strictly focused on this kind of change (ie: don't include any code reformatting); keep the PR/changes strictly on topic.

I might not be able to merge/approve it tonight; I'm just about to turn in for the night.

PR #529, opened 2024-02-09
Docs: Add inclusive / exclusive number ranges documentation for Randomizer

Mitchman215:
## Summary

Address #528 by adding information in `Randomizer.cs`'s xml param docs about whether the passed-in number ranges are inclusive or exclusive.

## Notes

All methods except for `Randomizer.Double()` and `Randomizer.Decimal()` had fully inclusive number ranges. Technically, `Randomizer.Float()` is also inclusive despite using `Randomizer.Double()` due to the rounding of `Convert.ToSingle()` ([fiddle](https://dotnetfiddle.net/QIEh0U))

### DIFF
--- a/Source/Bogus/Randomizer.cs
+++ b/Source/Bogus/Randomizer.cs
@@ -173,8 +173,8 @@ public class Randomizer
    /// <summary>
    /// Get a random double, between 0.0 and 1.0.
    /// </summary>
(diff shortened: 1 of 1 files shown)


bchavez:
Nice work @Mitchman215 ; thank you for the contribution!

PR #546, opened 2024-04-25
Fix issue #538 - invalid Costa Rican IBAN format

### LINKED ISSUE #538: Invalid Iban generation for CR (Costa Rica)
### Bogus NuGet Package

35.4.0

### .NET Version

.NET 6.0

### Visual Studio Version

17.8.4

### What operating system are you using?

Windows

en

### Problem Description

correct iban format for CR -> CR05 0152 0200 1026 2840 66 (example)
Bogus library is generating -> CR37 0051 0009 0045 0014 1 (example)

### LINQPad Example or Reproduction Steps

var iban = new Faker().Finance.Iban(false, "CR");

### Expected Behavior

expected valid 22 symbols iban

### Actual Behavior

invalid 21 symbols iban

### Known Workarounds

_No response_

### Could you help with a pull-request?

No

dangerman:
Fixes #538 

This PR updates the CR IBAN format, such that it has 22 characters instead of 21.
The fix should match the equivalent in faker-js (see https://github.com/faker-js/faker/pull/646).

We can compare the output of:
```faker.Finance.Iban(formatted:true, countryCode: "CR")```
on a validation site like https://wise.com/us/iban/checker

|Before | After |
| -- | -- |
| CR04 5521 6018 0960 1003 8 | CR80 6317 4801 1026 7666 76 |
|iban_validation_before|iban_validation_after|

The results are the same as with faker-js, so it's probably good enough.

### Note

The **Arabic locale data** [added in a previous PR](https://github.com/bchavez/Bogus/pull/149) appears to be different to what's in the checked out branch of the faker-js submodule.\
As a result, `gulp importLocales` gives us older (I think) locale data that fails the snapshot tests.
Anyway, I've left this out as it's out of scope and won't be an issue anyway when we have faker-js v8 parity.

### CODE REVIEW

**Comment by bchavez** [comment]
Excellent and amazing work @dangerman ;

Independently checked PR vs my own local changes, and hashes match:
```
    File: es.locale.bson
  CRC-32: 1620c28d
     MD5: 331c28818862d2ba39185e40a0426d67
   SHA-1: 0720f4e17a0efbe3df44dfe5a760296d6cbcade5
 SHA-256: 11ea02a589a98a834a2f8a8b6e90a30c28b210add38c0715cddadf793c3f1c70
 SHA-512: 73b6085587e999d05c2e3761ba3947976a01ecb0316d6b00dfd70a73f383d8ad3cb80b2dd9d4a72e72e7d84031114707d94288586a37f2cc334e4900b58cbb67
```

I should have a release out shortly.

PR #549, opened 2024-05-12
Correct the description of Random.ULong in README.md

logiclrd:
Just noticed that README.md has a copy/paste error, stating that `Randomizer`'s `ULong` method returns a value between -128 and 127, like `SByte`. The XML docs in the code do not have this error, just the README.


bchavez:
Cool! nice find. thank you very much! 👍 😸

Issue #555, opened 2024-07-19
LoremPixel service does not exist anymore and should be removed


marijnz0r:
### Bogus NuGet Package

v35.5.1 build 1436

### .NET Version

.NET 8

### Visual Studio Version

17.8

### What operating system are you using?

Windows

### What locale are you using with Bogus?

en_GB

### Problem Description

https://lorempixel.com/ does not exist anymore.
Suggestion: remove it from the codebase as it is obsolete for a long time already.

### LINQPad Example or Reproduction Steps

-

### Expected Behavior

-

### Actual Behavior

-

### Could you help with a pull-request?

Yes

bchavez:
Hi @marijnz0r , thank you for the PR. A new release has been made with your changes:
* https://www.nuget.org/packages/Bogus/35.6.0

Issue #418, opened 2022-04-04
picsum requests are outdated


21r8390:
### Version Information
| Software                       | Version(s) |
| ------------------------| ---------- |
| Bogus NuGet Package  |                 |

### What is the expected behavior?

Expected is that Bogus will create a link based on the [picsum.photos](https://picsum.photos) documentation.

```powershell
https://picsum.photos/id/69/640/480
``` 

### What is the actual behavior?

A link is created with `?image=id`, which is outdated. It is possible that this request will not be possible anymore at some point. 

```powershell
https://picsum.photos/640/480/?image=69
```

### Any possible solutions?

Specify the `Id` via the ULR and not as a parameter:
> (Important: grayscale should be specified via a parameter, however, according to the docu...)

```csharp
        public string PicsumUrl(int width = 640, int height = 480, bool grayscale = false, bool blur = false, int? imageId = null)
        {
            const string Url = "https://picsum.photos";

            var sb = new StringBuilder(Url);

            var n = imageId ?? this.Random.Number(0, 1084);
            sb.Append($"/id/{n}");

            sb.Append($"/{width}/{height}");

            if (grayscale)
            {
                sb.Append("?grayscale");
            }

            if (blur)
            {
                sb.Append(grayscale ? "&" : "?");
                sb.Append("blur");
            }

            return sb.ToString();
        }
```

### How do you reproduce the issue?

Create a picsum link via Bogus.

### Can you identify the location in Bogus' source code where the problem exists?
[DataSet/Images.cs](https://github.com/bchavez/Bogus/blob/b9049abf8b40203c09079741bcb328da95899f81/Source/Bogus/DataSets/Images.cs#L92)

```csharp
sb.Append($"/?image={n}");
```

### If the bug is confirmed, would you be willing to submit a PR?

No, because my recommended solution can already be found here in this report. Thus, this only needs to be added / overwritten.

bchavez:
I think we'll need to remove the LoremPixel service in our next major update.

prasadtelkikar:
I will take this, will raise PR soon

marijnz0r:
@prasadtelkikar Have you found time to create the pull request to remove lorempixel.com from the codebase?
@bchavez I would like to contribute by removing lorempixel.com from the codebase altogether. Would this be appreciated?

bchavez:
Hi @marijnz0r, yes I think removing LoremPixel.com is okay. I think the URLs and site have been broken for a long time.

Just a few more points:

[...]

PR #556, opened 2024-07-19
Remove LoremPixel image service

marijnz0r:
Solves  #555


bchavez:
Also, referencing issue #418; which this PR also resolves/fixes.

PR #570, opened 2024-11-02
Added support for Vietnamese Citizen Identity Card Number

git03-Nguyen:
Hello, this PR added extension method to generate Vietnamese 12-digit Citizen Identity Card Number (_"Căn cước công dân - CCCD"_ in local language):
- Add extension method `Cccd()` on `Person` object in `Bogus.Extensions.Vietnam` namespace.
- Add unit tests.
- Modify `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,6 +522,8 @@ In the examples above, all three alternative styles of using **Bogus** produce t
 * **`using Bogus.Extensions.UnitedStates;`**
 	* `Bogus.Person.Ssn()` - Social Security Number
 	* `Bogus.DataSets.Company.Ein()` - Employer Identification Number
+* **`using Bogus.Extensions.Vietnam;`**
+  * `Bogus.Person.Cccd()` - Vietnamese Citizen Identity Card Number (Căn Cước Công Dân - CCCD)
 * **`using Bogus.Distributions.Gaussian;`**
     * `Randomizer.GaussianInt()` - Generate an `int` based on a specific normal distribution.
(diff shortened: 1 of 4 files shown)


bchavez:
Great work. Thank you for this pull request. I'll try to get to this soon, but might be a little difficult with the holidays coming up, I'll be at the airport in a few days and hopefully I will try to review this while I wait at the terminal

git03-Nguyen:
> Great work. Thank you for this pull request. I'll try to get to this soon, but might be a little difficult with the holidays coming up, I'll be at the airport in a few days and hopefully I will try to review this while I wait at the terminal

Happy to hear that ^^. It's just some simple work done in my spare time cuz I really love Bogus.
Anw, tell me if the code need any changes.

bchavez:
Hi @git03-Nguyen; thanks again for the PR.

After reviewing this PR at the airport, it appears we are slightly misaligned with some of city names which could potentially throw some exceptions when this code executes.

I diffed the following `vi` locale city names from our JSON:
* https://github.com/bchavez/Bogus/blob/905a7e753ed2d7bfd42fc5099a50278adf45b467/Source/Bogus/data/vi.locale.json#L36

then compared it to the `CityCodes` in this PR, and we have a few city names that look like they might not match; 

image

image

We should probably get these two lists corrected and synchronized so we don't stumble on an exception with `throw new ArgumentException($"City {p.Address.City} is not supported."` when `Person.Cccd()` executes.

Let me know what you think.

PR #590, opened 2025-02-21
implicit usings

SimonCropp:
I like the feature. but many dont. just burn it if u dont like the approach


bchavez:
Thanks for this, but like you mention, not too much of a huge fan of implicit usings.

I've been burned a few times of magical usings that pick up on the wrong type that I intended. I would take a PR however, for explicit disablement in csproj. 👍

Issue #565, opened 2024-09-24
Add Vehicle, Music, and Person to Parse method


Perks-of-Being-a-Cauliflower:
### Description

I am using Faker dynamically and it would be helpful if the Vehicle, Music, and Person methods were added to the Parse method. Vehicle and Music are simple, they just need to be added to the list of MustsacheMethods passed into the Parse method. Person appears to be a bit more complex. When it is added to Parse it seems to mess up the seed which causes some of the randomizer parse tests to fail. I assume this is because Person takes Randomizer as a parameter itself.

I have attempted the changes on a feature branch with Vehicle and Music working fine but seems a bit redundant if I can't get Person working. Person would also require methods for each of its properties so they can be registered as MustacheMethods.

### LINQPad Code Example

Bogus Version="35.6.1"

```
using Bogus;

void Main()
{
    var faker = new Faker();
    var manufacturer = faker.Parse("{{vehicle.manufacturer}}");
    var genre = faker.Parse("{{music.genre}}");
    var firstname = faker.Parse("{{person.getfirstname}}");
    var lastname = faker.Parse("{{person.getlastname}}");
    var fullname = faker.Parse("{{person.getfullname}}");
	
    if (manufacturer == null)
    {
        throw new Exception("Parse failed manufacturer is null");
    }
	
    if (genre == null)
    {
        throw new Exception("Parse failed genre is null");
    }
	
    if (firstname == null || lastname == null || fullname == null)
    {
        throw new Exception("Parse failed Person values are null");
    }
	
    if ($"{firstname} {lastname}" != fullname)
    {
        throw new Exception("Person values not generated correctly with Parse");
    }
}
```

### What alternatives have you considered?

For Person I tried using a switch statement but this is messy and not ideal for my use case.

### Could you help with a pull-request?

Yes

Perks-of-Being-a-Cauliflower:
I managed to implement a suitable fix. Created a pull request and am just waiting for it to be approved.

PR #602, opened 2025-03-29
Add Vehicle, Music, and Person to Parse method #565

Perks-of-Being-a-Cauliflower:
Faker now supports Music, Vehicle, and Person for the Parse method. added some unit tests to check that values are parsed correctly. In order for Person to be used with Parse, methods were added to Person class to facilitate Mustache methods.

Needed to create a second initialisation for Person in Faker class because Populate() was being called by default which was changing all the expected values for many unit tests. It would be possible to change the existing tests to accept the new values however, I think conditionally populating is a more elegant solution.

[Issue](https://github.com/bchavez/Bogus/issues/565)

### DIFF
--- a/Source/Bogus.Tests/HandlebarsTests/ArgumentsTest.cs
+++ b/Source/Bogus.Tests/HandlebarsTests/ArgumentsTest.cs
@@ -137,4 +137,121 @@ public class ArgumentsTest : SeededTest
       faker.Parse("{{date.timespan(00:00:25)}}")
          .Should().Be("00:00:15.0880571");
    }
+
+
+   [Fact]
+   public void can_parse_vehicle()
+   {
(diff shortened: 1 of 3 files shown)


Perks-of-Being-a-Cauliflower:
In my eagerness to add the functionality to complete the Parse method, I left in some unnecessary code. Hopefully my implementation is a bit cleaner now.

Perks-of-Being-a-Cauliflower:
I noticed that some more recent pull requests have been looked at. Just want to check if there is anything I have missed or need to do for mine. Also happy to try and rework the code if there is an issue with it. Its a relatively small change but I think it could be valuable for people like myself who utilize the parse method.

Love the library. It's such an amazing tool.

PR #604, opened 2025-04-15
Refactored person.cs

killacoug9:
Added better documentation to the Person.cs internal contructor. Added Tests for the person contructors, person interactions with randomizer, and internal constructor tests using reflection. Refactored Person.cs. Fixed member ordering to comply with C# coding conventions and best practices. Added some documentation to clarify method. Performed method extraction in contructors to improve maintainability. Made public fields into properties for improved data control.

### DIFF
--- a/Source/Bogus.Tests/PersonTest.cs
+++ b/Source/Bogus.Tests/PersonTest.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using Bogus.DataSets;
 using Bogus.Extensions.Brazil;
 using Bogus.Extensions.Canada;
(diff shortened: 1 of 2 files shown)


bchavez:
Hi. The change set in this PR is too large for review. **You are changing too much... also, for what purpose? What is the goal?** Consider reading the contributing guidelines here:

* https://github.com/bchavez/Bogus/blob/master/CONTRIBUTING.md

Additionally, this repository has over 100 million downloads on NuGet, and we need to be pragmatic with code changes to this repository. That is, I don't necessarily accept or change code for the sake of changing code or because a guideline or best practice says so. Also, by who's C# standards? Microsoft's? or the person maintaining this repository? IMHO, it's okay to bend the rules sometimes.

In this PR, we are:
* Changing public fields to public properties.
  * Again, for what pragmatic purpose? Do you (or I) need improved "data control" on these public fields? I don't. I'm sure someone out there has taken a dependency on these fields in one way or another, maybe through serialization; and we are changing these public fields to public properties, potentially making a breaking change for some consumers downstream. Is our response to the broken consumer of this library "because it improves data control" that we have no need for? I dunno. It's not bothering anyone, maybe we leave it alone. Public field vs public property in this usage scenario is practically the same thing. Sure, maybe public properties are the more practical choice when this class was originally designed, but there isn't a strong reason to change it now and potentially break someone's app or ruin someone's work day.
* Changing coding style by removing `this.`. **Do not change the coding style in this file.**
* Using reflection when there is no reason to. Why are we over engineering a solution to call an internal constructor in `Bogus.Tests`? The main assembly already has `[InternalsVisibleTo]`; https://github.com/bchavez/Bogus/blob/2e84969c556058437bd4ac9dce2edc7209339e7f/Source/Bogus/Properties/AssemblyInfo.cs#L15
  Please refer to Microsoft's documentation for more information on how `[InternalsVisibleTo]` works.

Unfortunately, I won't accept this PR. For future reference, it might be better to first submit a GH issue first, before working on a PR, where we can discuss any proposed changes to avoid throwing away code.

PR #605, opened 2025-04-15
Refactored Randomizer.cs

fruitslayer1:
Randomizer.cs:
- Added or improved documentation for several methods and attributes
- Reordered members and changed capitalization on variable names to fit conventional C# best practices
- Encapsulated Seed attribute
- Added additional overload for Number method and removed redundant default parameter values
- Replaced multiple declarations of explicit typing with 'var' keyword
- Removed redundant usage of 'this' keyword
- Placed attributes 'AlphaChars' and 'HexChars' as local variables inside their respective methods
- Replaced 'if null' branches and some assignments with assignment operators
- Inversed 2nd 'if' statement in ClampString() method
- Changed Uuid() method to simply call Guid() instead
- Removed redundant 'if-else' statement in EnumValues<T>()
- Removed redundant 'else' statement and initialization of variables in Utf16String()

RandomizerTest.cs: Added test for ClampString() method
BelgianExtensionTests.cs: Added explicit skip for can_generate_national_number_for_belgium() test

### DIFF
--- a/Source/Bogus.Tests/ExtensionTests/BelgianExtensionTests.cs
+++ b/Source/Bogus.Tests/ExtensionTests/BelgianExtensionTests.cs
@@ -15,7 +15,7 @@ public class BelgianExtensionTests : SeededTest
       _faker = new Faker();
    }
 
-   [Fact]
+   [Fact(Skip = "Explicit")]
    public void can_generate_national_number_for_belgium()
    {
       // Act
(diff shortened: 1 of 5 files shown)


bchavez:
I'm sorry, but the change set is way too large for review. Especially for a core piece of code like `Randomizer.cs`; closing this PR for roughly the same reason as https://github.com/bchavez/Bogus/pull/604

PR #613, opened 2025-07-25
Added Australian registration

licon4812:
This pull request introduces a new feature to generate Australian vehicle registration plates with state-specific formats. It includes both the implementation of the feature and corresponding unit tests to ensure correctness. The most important changes are grouped into feature implementation and testing.

### Feature Implementation:

* Added a new extension method `AusCarRegistrationPlate` in `ExtensionsForAustraliaRegistrationPlate.cs` to generate Australian vehicle registration plates based on the state, year, and date range. This method validates the state, determines the registration date, and applies state-specific formatting rules.
* Defined `PlateFormatRule` and `StatePlateRules` to encapsulate the logic for state-specific plate formats based on year ranges, including support for custom generation logic.
* Implemented helper methods such as `GenerateRegistrationDate`, `GenerateRegistrationPlate`, and `GeneratePlateFromFormat` to handle random date generation, format selection, and plate string construction.

### Testing:

* Added a new test class `AustraliaExtensionTests` in `AustraliaExtensionTests.cs` containing unit tests to verify the correctness of the `AusCarRegistrationPlate` method for various states and year ranges.
* Included a parameterized test `Generates_Correct_Plate_Format_For_State_And_Year` to validate that the generated plate matches the expected format for a given state and year.
* Added a test `Generates_Correct_Plate_Format_For_Random_State` to ensure that the method works correctly when no specific state is provided, generating a valid plate for a random state.

### DIFF
--- a/Source/Bogus/Extensions/Australia/ExtensionsForAustraliaRegistrationPlate.cs
+++ b/Source/Bogus/Extensions/Australia/ExtensionsForAustraliaRegistrationPlate.cs
@@ -33,9 +33,10 @@ public static class ExtensionsForAustraliaRegistrationPlate
    /// Different vehicle types also have different registration plate formats.
    /// Currently, this method only supports the "Car" or universal type of vehicle registration plates. 
    /// </remarks>
-   public static string AusCarRegistrationPlate(this Vehicle vehicle, DateTime from, DateTime to, string state)
+   public static string AusCarRegistrationPlate(this Vehicle vehicle, DateTime from, DateTime to, string state = null)
    {
-      if (! SupportedStates.Contains(state)) 
+      state ??= GenerateRandomState();
+      if (!SupportedStates.Contains(state))
          throw new ArgumentException($"Unsupported Australian state: {state}. Supported states are: {string.Join(", ", SupportedStates)}", nameof(state));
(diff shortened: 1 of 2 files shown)


Source/Bogus/Extensions/Australia/ExtensionsForAustraliaRegistrationPlate.cs
```diff
@@ -0,0 +1,175 @@
...
+            default: sb.Append(c); break;
+         }
+      }
+      return sb.ToString();
+   }
+
+   private static string GenerateRandomState()
+   {
+      var randomizer = new Randomizer();
+      return randomizer.ListItem(SupportedStates.ToList());
```
bchavez, line 173: I don't think we should be creating a `new Randomizer()` here. Try pulling the `vehicle.Random` instead to pull a randomizer that has already been instantiated. We need generally need to re-use the randomizer that has already been created at start up to preserve determinsitc results.

bchavez: See this for example:
* https://github.com/bchavez/Bogus/blob/2e84969c556058437bd4ac9dce2edc7209339e7f/Source/Bogus/Extensions/UnitedKingdom/ExtensionsForGreatBritainRegistrationPlate.cs#L150

PR #614, opened 2025-07-28
Add extension method for generating Iranian national number

amirh-p:
This pull request introduces a new feature to generate Iranian national number. The unit tests are also included.

### Feature Implementation
- Added a new static class called `ExtensionsForIran` inside a new folder called `Iran` in the Extensions folder.
- Added a new extension method inside the `ExtensionsForIran` class called `IranianNationalNumber` to generate valid Iranian national number.
### Feature Testing
- Added a new class called `IranianExtensionTests` inside the ExtensionTests folder. This class contains four newly added methods described below 👇
1. `ValidateIranianNationalNumber` : This is responsible for validating any given national number via a specific algorithm and is used by other test methods.
2. `can_create_valid_iranian_national_number` : This will ensure the output matches the expected pattern and is actually a valid national number.
3. `iranian_national_number_generated_twice_are_equal` : This will ensure that national number remains the same for the `Person` object even if generated twice.
4. `iranian_national_number_should_be_different_for_different_persons` : This will ensure the national number of two different `Person` objects are not the same.

### DIFF
--- a/Source/Bogus.Tests/ExtensionTests/IranianExtensionTests.cs
+++ b/Source/Bogus.Tests/ExtensionTests/IranianExtensionTests.cs
@@ -7,7 +7,6 @@ namespace Bogus.Tests.ExtensionTests;
 
 public class IranianExtensionTests : SeededTest
 {
-
    [Fact]
    public void can_create_valid_iranian_national_number()
    {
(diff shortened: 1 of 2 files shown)


bchavez:
LGTM. Thank you.

PR #615, opened 2025-08-12
fix incorrect as usages

SimonCropp:
places where if it isnt that type, it would result in a null ref. so in the case where the type assumption is wrong, it is better to get a helpful exception about incorrect type, than a null ref

### DIFF
--- a/Source/Bogus/Bson/Bson.cs
+++ b/Source/Bogus/Bson/Bson.cs
@@ -189,12 +189,12 @@ public class Bson
          case BValueType.Object:
             ms.WriteByte(0x03);
             EncodeCString(ms, name);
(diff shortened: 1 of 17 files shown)


bchavez:
Going to hold on this one; and get a new release out; and then merge this after vNext.

This one is a bit more tricky; not 100% sure yet if `null` might be checked by methods receiving or sending these values.

bchavez:
actually sorry, the i think the cast would work even if the value was null; sorry long day;

I don't think this actually changes or makes the exception happen earlier; ie `var x = (string)null;` just works, but I think i see your point.

Fool256 (approved):

PR #621, opened 2025-09-22
Extensions for the Netherlands

JesseKlaasse:
- Add Dutch vehicle registration plates (kentekenplaten)
- Correct Dutch zip codes and phone numbers


bchavez (changes requested):

JesseKlaasse:
@bchavez You could check again, I have made some changes.

Source/Bogus/data/nl.locale.json
```diff
@@ -2857,7 +2857,15 @@
...
-      "#### ??"
+      "1### ??",
+      "2### ??",
+      "3### ??",
+      "4### ??",
+      "5### ??",
+      "6### ??",
+      "7### ??",
+      "8### ??",
+      "9### ??"
```
bchavez, line 2868: we shouldn't be modifying the underlying `data` set directly; any changes here will get clobbered over; if you need to update a dataset you need to have this JSON node replaced by `data_extend`/`nl` file; `data_extend` gets merged with faker.js; so ultimately any changes here will get clobbered over by an update process from faker.js.

this `/data/nl.locale.json` file is the ultimate output of the merge process between Bogus `data_extend` and faker.js import.

you can learn more about how faker.js data is imported into Bogus here:
* https://www.youtube.com/watch?v=KoEOsfrRUqk&t=1682s

image

bchavez (changes requested):

PR #622, opened 2025-10-25
Adds music and vehicle faker classes to parse method and unit tests

Perks-of-Being-a-Cauliflower:
These are some small changes taken from my other branch/pull request  "Add Vehicle, Music, and Person to Parse method #602". Since adding the Person class is likely to be more involved, I split the changes as requested.

### Issue/Feature

Faker contains a parse method however it is missing implementations for Vehicle and Music.

### Implimentation

Vehicle:

    Added to list of Mustache Methods passed into Parse method
    Added unit tests to check that values can be returned from parse

Music

    Added to list of Mustache Methods passed into Parse method
    Added unit tests to check that values can be returned from parse


bchavez:
Fixed in v35.6.5: https://www.nuget.org/packages/Bogus/35.6.5

PR #624, opened 2025-11-01
Issue #565. Provides alternate solution to pull request #602. Adds pe…

Perks-of-Being-a-Cauliflower:
Issue #565. 
Provides alternate solution to pull request #602. 
Adds person class by modifying tokenizer to use class properties getters as methods.
It's a little bit messy because of limitations of older supported .NET versions.
Requirements:
- No extra state to manage.
- Easily expanded if Person got more properties or if another class needed its properties added.
- Utilize similar design as existing method registration to avoid creating deviations in process.
- Modify as little of the existing code as possible to avoid introducing unforeseen bugs in existing functionality.
- Avoid redefining existing classes/functionality to avoid overhead.

### DIFF
--- a/Source/Bogus.Tests/HandlebarsTests/ArgumentsTest.cs
+++ b/Source/Bogus.Tests/HandlebarsTests/ArgumentsTest.cs
@@ -137,4 +137,90 @@ public class ArgumentsTest : SeededTest
       faker.Parse("{{date.timespan(00:00:25)}}")
          .Should().Be("00:00:15.0880571");
    }
+
+   [Fact]
+   public void can_parse_person()
+   {
+      var faker = new Faker();
+      var firstname = faker.Parse("{{person.firstname}}");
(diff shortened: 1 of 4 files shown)


Perks-of-Being-a-Cauliflower:
Hi @bchavez 
Hopefully this is more what you had in mind. I feel like my code could maybe be a bit cleaner, but I had to fight against the older .NET versions and some other weirdness. At the very least it should act as a proof of concept should this be the route you want to go down.

Perks-of-Being-a-Cauliflower:
Hi @bchavez
Would like to get your thoughts on this implementation. I have a feeling that this attempt wasn't exactly what you had in mind but it doesn't add extra context to track and seems to meld decently into the existing parse capability.

I'm happy to modify it if you think there are issues with this implementation. I've tried my best with both this and my previous PR not to add any additional overhead or debt but if there is anything that doesn't sit right just let me know.

Cheers

PR #633, opened 2026-03-17
Fix indeterministic behavior of CodiceFiscaleGenerator

### LINKED ISSUE #632: CodiceFiscale generator is not deterministic
### Bogus NuGet Package

35.6.5

### .NET Version

.NET 10

### Visual Studio Version

_No response_

### What operating system are you using?

Windows

it

### Problem Description

Given a static seed CodiceFiscale should generate a deterministic code, like other generators, but it doesn't. It depends on GetHashCode(), which itself is not deterministic.

### LINQPad Example or Reproduction Steps

```csharp
public sealed class Player
{
    public required Guid Id { get; set; }
    public required string FirstName { get; set; }
    public required string LastName { get; set; }
    public required string TaxCode { get; set; }
}

Randomizer.Seed = new Random(9612867);

var testPlayers = new Faker<Player>("it")
    .StrictMode(true)
    .RuleFor(x => x.Id, f => f.Random.Guid())
    .RuleFor(x => x.FirstName, f => f.Name.FirstName())
    .RuleFor(x => x.LastName, f => f.Name.LastName())
    .RuleFor(x => x.TaxCode, f => f.Person.CodiceFiscale());

var players = testPlayers.Generate(2);
```

### Expected Behavior

Every time I run the test using this code it should produce the exact same codes.

### Actual Behavior

Every run Bogus generates a different codes.

### Known Workarounds

_No response_

### Could you help with a pull-request?

Yes

Mephistofeles:
Replace .NET GetHashCode with a deterministic FNV-1a hash for city code in ExtensionsForItaly.CodiceFiscale.cs to ensure cross-runtime stability.

Fixes https://github.com/bchavez/Bogus/issues/632

### DIFF
--- a/Source/Bogus/Extensions/Italy/ExtensionsForItaly.CodiceFiscale.cs
+++ b/Source/Bogus/Extensions/Italy/ExtensionsForItaly.CodiceFiscale.cs
@@ -132,14 +132,44 @@ internal static class CodiceFiscaleGenerator
       //To guarantee code stability for a person, we generate
       //fake city-of-birth code through surname and birth date.
       //Using actual city code database would be too heavy.
-      sb.Append(lastName[0].ToString().ToUpper());
-      var birthDatePositiveHash = Math.Abs(birthday.GetHashCode());
-      sb.Append((birthDatePositiveHash % 1000).ToString("000"));
+      var cityHash = ComputeStableHash(lastName, firstName, birthday, male);
+      sb.Append((char)('A' + cityHash % 26));
+      sb.Append((cityHash / 26 % 1000).ToString("000"));
 
       var checksum = ComputeChecksumCodiceFiscale(sb.ToString(), validChecksum);
       sb.Append(checksum);
 
       return sb.ToString();
+   }
+
+   /// <summary>
+   /// Computes a deterministic, runtime-stable hash from all fiscal code inputs using FNV-1a.
+   /// Unlike <see cref="object.GetHashCode"/>, this produces the same value on every .NET runtime and version.
+   /// </summary>
+   /// <param name="lastName">Last name of the holder</param>
+   /// <param name="firstName">First name of the holder</param>
+   /// <param name="birthday">Birthday of the holder</param>
+   /// <param name="male">Indicates whether the holder is male</param>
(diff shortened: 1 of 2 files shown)


Source/Bogus/Extensions/Italy/ExtensionsForItaly.CodiceFiscale.cs
```diff
@@ -132,16 +132,46 @@ public static string Generate(
...
+
+         foreach (var c in lastName.ToUpperInvariant())  { hash = (hash ^ c) * prime; }
+         foreach (var c in firstName.ToUpperInvariant()) { hash = (hash ^ c) * prime; }
+         hash = (hash ^ (uint)birthday.Year) * prime;
+         hash = (hash ^ (uint)birthday.Month) * prime;
+         hash = (hash ^ (uint)birthday.Day) * prime;
+         hash = (hash ^ (male ? 1u : 0u)) * prime;
+
+         return hash;
+      }
```
bchavez, line 172: Is it possible to have an implementation where we are not using `unchecked`? I would generally prefer to have an implementation that doesn't require an `unchecked` block.

bchavez (changes requested):
Good find. Thank you for bringing up the non-deterministic `.GetHashCode()`  usage. Most definitely we want to remove non-deterministic calls and replace them with deterministic method calls as much as possible.

Source/Bogus/Extensions/Italy/ExtensionsForItaly.CodiceFiscale.cs
Mephistofeles: The FNV-1a is a simple algorithm, but it depends on the overflow wrapping.
I thought about using one of the algorithms in the System.IO.Hashing namespace, but it requires .NET 4.6.2+.
I guess I can use for example SHA256 if you don't want `unchecked ` in the code.

bchavez: Assuming there's a comparable implementation in System.IO.Hashing and an
open source version in .NET, I suppose we could also copy the BCL
implementation from that namespace for our usages here.

